PR #2, opened 2017-12-15
add amp runtime api

sebastianbenz:


runtime-version/lib/MaxAge.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  /**
+   * Parses a max-age header string and returns a new MaxAge instance.
+   *
+   * @param {string} str a max-age header string
+   * @returns {MaxAge}
+   */
+  static parse(string) {
+    const match = string.match(/max-age=([0-9]+)[^0-9]?/i);
```
andreban: Can we add a check for null to string? eg: `if (!string) return MaxAge.zero();`

runtime-version/lib/MaxAge.js
```diff
@@ -0,0 +1,67 @@
...
+    let maxAge = 0;
+    if (match) {
+      maxAge = parseInt(match[1], 10);
+    }
+    return new MaxAge(Date.now(), maxAge);
```
andreban: What's effectively the difference between this and MaxAge.zero()? Could it be worth to create a MaxAge.now(), and re-use it when checking for the null string check above?

runtime-version/.eslintrc.json
```diff
@@ -0,0 +1,37 @@
+{
+  "extends": "google",
+  "installedESLint": true,
+  // http://eslint.org/docs/rules/
```
andreban: nit: Comments aren't allowed in JSON. A common solution is to use a "_comment" field to add the comment and not end up with an invalid JSON. ESLint doesn't seem to care about this though.

runtime-version/lib/MaxAge.js
sebastianbenz: made this more explicit

runtime-version/lib/MaxAge.js
sebastianbenz: done

runtime-version/lib/RuntimeVersion.js
```diff
@@ -0,0 +1,83 @@
...
+   */
+  currentVersion(options = {}) {
+    if (options.canary) {
+      return this.getAndUpdateCanary_();
+    }
+    return this.getAndUpdateRelease_();
+  }
+
+  /* PRIVATE */
+  getAndUpdateRelease_() {
```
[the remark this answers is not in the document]
sebastianbenz: let's standardise on postfixing

sebastianbenz:
PTAL

runtime-version/lib/RuntimeVersion.js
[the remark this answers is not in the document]
andreban: SGTM

PR #4, opened 2018-01-03
Enable CI tests on Travis

ithinkihaveacat:
To get this working, first go to https://travis-ci.org/ampproject/amp-toolbox and enable Travis for the repo. I think it will then be possibly to manually trigger a build for this PR (?). (Once merged, this will happen automatically for all subsequent PRs.)

The renaming commits are to change `AmpBoilerPlateTransformer` to `AmpBoilerplateTransformer` so that tests work on case sensitive filesystems. (Seems like it needs to be done in stages to avoid confusing git on macOS.) 

([Example build](https://travis-ci.org/ithinkihaveacat/amp-toolbox).)


.travis.yml
```diff
@@ -0,0 +1,9 @@
+language: node_js
+node_js:
+  - "node"
+  - "6"
+  - "8"
+install:
+  - for d in runtime-version ssr ; do cd $d ; npm install || exit ; cd - ; done
```
ithinkihaveacat: Can also generate list of directories that contain `package.json` via `find . -depth 2 -name 'package.json' -exec dirname \{\} \;` but it's maybe a bit magical/weird when there's only two to iterate through. (Another option is just to copy and paste the npm install/test lines, which will also avoid the need for the strange `|| exit`.)

sebastianbenz: How about adding a script in the root project that does this?

ithinkihaveacat: That's possible, how about `.travis/install` and `.travis/script` (to match the keys)? I don't think there's a convention, unfortunately.

ithinkihaveacat: Now moved to `.travis/{install,script}`. `find` doesn't work because not cross platform enough but the shell script equivalent is okay.

sebastianbenz:
That's fantastic! Thanks Michael!

sebastianbenz:
Can I merge?

ithinkihaveacat:
@sebastianbenz Yes, it's ready to merge--thanks!

PR #5, opened 2018-01-04
Adds "cheerio" demo

ithinkihaveacat:
This example demonstrates the use of [Cheerio](https://github.com/cheeriojs/cheerio) to manipulate the DOM in a jQuery-like way, as an alternative to the [domutils](https://github.com/fb55/domutils)-style transformations of the built-in transformers.

e.g. `$('title').text('SSR: ' + $('title').text())` prepends `SSR` to the `<title>`.

### DIFF
--- a/ssr/demo/cheerio/README.md
+++ b/ssr/demo/cheerio/README.md
@@ -4,7 +4,7 @@ This example demonstrates the use of
 [Cheerio](https://github.com/cheeriojs/cheerio) to apply jQuery-like
 transformations to a DOM, as an alternative to the
 [domutils](https://github.com/fb55/domutils)-style transformations of the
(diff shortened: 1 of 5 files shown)


sebastianbenz:
That's very cool! I wasn't aware that cheerio understands the same node model.

ithinkihaveacat:
Thanks for merging! Spent several hours deep in the loading code of cheerio and parse5 but was very happy to discover that `cheerio.load(tree.root.children)` was sufficient in the end.

(Have been looking at this as part of a project to rewrite https://github.com/ithinkihaveacat/amptimizer/blob/beebo/index.ts using `amp-toolbox`, BTW.)

PR #16, opened 2018-01-19
Inline v0.css 2

sebastianbenz:


ssr/lib/transformers/AmpBoilerplateTransformer.js
```diff
@@ -51,16 +61,30 @@ class AmpBoilerplateTransformer {
     return null;
   }
 
-  _addStaticCss(tree, node) {
+  _addStaticCss(tree, node, params) {
+    if (params.ampRuntimeVersion) {
```
andreban: Does it make sense to have an specific configuration to inline / link the CSS? Inlining the CSS means one less request on the first view, given that the CSS hasn't been cached already. Linking decreases the HTML size, and improving data usage once the file is cached. It might be good to have it as a separate option.

sebastianbenz: It only makes sense if the AMP runtime is versioned - otherwise you might run into inconsistencies between v0.js and inlined v0.css

andreban: Inlining only makes sense, if it's versioned. But linking may still make sense when versioned.

sebastianbenz: sorry, missed that - yeah that makes sense

sebastianbenz: What should the default be? inline or link?

andreban: Inlining performs better on WPT. So, that sounds like a good default. 

In the future, we need to check real-user monitoring, compare both, and provide guidance on which one is better, and when. As an example, a site with a high rate of new users might be better on inlining, but a site skewed towards repeat users may be better with linking...

sebastianbenz: done - added an option

PR #21, opened 2018-01-22
Makes middleware compatible with Connect

andreban:
- Checking for static resources verifies existance of req.accepts
- Call res.write and res.end when transformation is finished,
  instead of calling res.send.
- Updated JSDoc

Change-Id: Ia6edf4e57c8779954b5173bc2664ef3d194e93aa

### DIFF
--- a/ssr-express/spec/lib/AmpSsrMiddlewareSpec.js
+++ b/ssr-express/spec/lib/AmpSsrMiddlewareSpec.js
@@ -61,15 +61,15 @@ describe('Express Middleware', () => {
         });
     });
 
-    it('Skips Resource Requests', () => {
-      const staticResources = ['/image.jpg', '/image.svg', '/script.js', '/style.css'];
-      const runStaticTest = url => {
-        runMiddlewareForUrl(middleware, url)
-          .then(result => {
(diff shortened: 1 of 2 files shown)


andreban:
Added tests

ssr-express/spec/lib/AmpSsrMiddlewareSpec.js
```diff
@@ -59,5 +60,30 @@ describe('Express Middleware', () => {
           expect(result).toEqual('original');
         });
     });
+
+    it('Skips Resource Requests', () => {
+      const staticResources = ['/image.jpg', '/image.svg', '/script.js', '/style.css'];
```
sebastianbenz: tip: you can dynamically generate specs:

```
['/image.jpg', '/image.svg', '/script.js', '/style.css'].forEach(url => {
  it(url, () => {
      runMiddlewareForUrl(middleware, url)
           .then(result => {
             expect(result).toEqual('original');
           });
  });
});
```

andreban: This is great. Rewrote.

PR #27, opened 2018-01-24
Enables middleware to work with runtimeVersion

andreban:


ssr-express/lib/AmpSsrMiddleware.js
```diff
@@ -97,10 +107,23 @@ class AmpSsrMiddleware {
         // This is a request for the canonical URL. Generate the AMP equivalent
         // in order to add it to the link rel tag.
         const linkRelAmpHtmlUrl = urlMapping.toAmpUrl(req.url);
-        ssr.transformHtml(body, {ampUrl: linkRelAmpHtmlUrl})
+
+        runtimeVersion()
+          .then(version => version)
```
sebastianbenz: Is this needed?

ssr-express/lib/AmpSsrMiddleware.js
```diff
@@ -97,10 +107,23 @@ class AmpSsrMiddleware {
...
+          .then(version => ssr.transformHtml(body, {
+            ampUrl: linkRelAmpHtmlUrl,
+            ampRuntimeVersion: version
+          }))
           .then(transformedBody => {
-            res.write(Buffer.from(transformedBody, 'utf-8'));
-            res.end();
+            res.status(200).send(transformedBody);
+          })
+          .catch(err => {
```
sebastianbenz: Do multiple catch handlers work in a promise chain?

ssr-express/lib/AmpSsrMiddleware.js
andreban: Actually, no - removed.

ssr-express/lib/AmpSsrMiddleware.js
andreban: They do!

sebastianbenz: Will the promise chain stop after the first handler?

andreban: The first `catch` handler is returning null, which will be passed to the `then` handler after it. So, the version will be null if retrieving the runtimeVersion fails.

If any error happens from then on, the second `catch` handler will be triggered.

andreban: The last test on the Spec is testing this behaviour.

sebastianbenz:
Are you planning to update the sample in this PR as well?

andreban:
Sorry, I missed the first comment. Do you think it's worth having a demo with the simple usage and another with all the custom options?

sebastianbenz:
Not sure If we need all custom options, but we should at least have one
with versioned runtime.

andreban:
Cool. I'll move the existing demo to demos/simple and create a demos/runtime-version. But I'd rather do it on another PR.

PR #29, opened 2018-01-26
Adds runtime-version demo for the Middleware

andreban:
- Fixes error where runtime treats paths ending as '.com/' as
  resources.

- Option to pass transformation parameters through the request.

- Adds demo/runtimeVersion.js, showing how to run the Middelware
  with a custom runtime.


ssr-express/demo/runtimeVersion.js
```diff
@@ -0,0 +1,45 @@
...
+const express = require('express');
+const path = require('path');
+const app = express();
+const AmpSsrMiddleware = require('../index.js');
+const runtimeVersion = require('amp-toolbox-runtime-version');
+
+// Wrap the current version in our own function. The function has no parameters and returns
+// a Promise<Number>.
+// This function will be called every time a new AMP is transformed.
+const currentVersion = () => runtimeVersion.currentVersion();
```
sebastianbenz: this is not needed, simply pass `runtimeVersion.currentVersion` as an option

andreban: Tried that, but it breaks with the following error:

```
TypeError: Cannot read property 'getAndUpdateRelease_' of undefined
```

PR #47, opened 2018-05-25
 Adds the `update-cache` module

andreban:
Implements update-cache, using cache-list and cache-url libraries.
Generates signature, as described on the AMP docs at
https://developers.google.com/amp/cache/update-ping#update-cache-request

### DIFF
--- a/update-cache/README.md
+++ b/update-cache/README.md
@@ -5,16 +5,16 @@ https://developers.google.com/amp/cache/update-ping#update-cache-request).
 
 ## Usage
 ```javascript
-  const createUpdateCacheUrlProvider = require('amp-toolbox-update-cache');
+  const UpdateCacheUrlProvider = require('amp-toolbox-update-cache');
 
   // Load the private key, generated as described on 
   // https://developers.google.com/amp/cache/update-cache#rsa-keys
   const privateKey = '...';
 
   // Create an instance of the factory using the private key.
-  const updateCacheUrlProvider = createUpdateCacheUrlProvider(privateKey);
+  const updateCacheUrlProvider = UpdateCacheUrlProvider.create(privateKey);
 
-  updateCacheUrlProvider.fromOriginUrl('https://www.example.com')
+  updateCacheUrlProvider.calculateFromOriginUrl('https://www.example.com')
     .then(cacheUpdateUrls => {
       cacheUpdateUrls.forEach(cacheUpdateUrlInfo => {
         console.log('Cache ID:' + cacheUpdateUrlInfo.cacheId);
--- a/update-cache/README.md
+++ b/update-cache/README.md
@@ -1,0 +1,25 @@
+#AMP-Toolbox update-cache
+
+Generates update-cache invalidation URLs, as described on the AMP documentation at 
+https://developers.google.com/amp/cache/update-ping#update-cache-request.
+
+## Usage
+```javascript
+  const createUpdateCacheUrlFactory = require('amp-toolbox-update-cache');
+
+  // Load the private key, generated as described on 
+  // https://developers.google.com/amp/cache/update-cache#rsa-keys
+  const privateKey = '...';
+
+  // Create an instance of the factory using the private key.
+  const updateCacheUrlFactory = createUpdateCacheUrlFactory(privateKey);
+  const timestamp = (Date.now() / 1000) | 0; // timestamp as a UNIX Epoch in seconds.
+
+  const cacheUpdateUrls = await updateCacheUrlFactory.fromOriginUrl('https://www.example.com');
+  cacheUpdateUrls.forEach(cacheUpdateUrlInfo => {
+    console.log('Cache ID:' + cacheUpdateUrlInfo.cacheId);
+    console.log('Cache Name:' + cacheUpdateUrlInfo.cacheName);
+    console.log('cache-update URL:' + cacheUpdateUrlInfo.updateCacheUrl);
+  });
+    
+```
--- a/update-cache/README.md
+++ b/update-cache/README.md
@@ -1,25 +1,25 @@
 #AMP-Toolbox update-cache
 
-Generates update-cache invalidation URLs, as described on the AMP documentation at 
-https://developers.google.com/amp/cache/update-ping#update-cache-request.
+Generates update-cache invalidation URLs, as described on the [Google AMP Cache documentation](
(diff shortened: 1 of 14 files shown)


update-cache/README.md
```diff
@@ -0,0 +1,25 @@
+#AMP-Toolbox update-cache
+
+Generates update-cache invalidation URLs, as described on the AMP documentation at 
+https://developers.google.com/amp/cache/update-ping#update-cache-request.
```
sebastianbenz: * can you turn this into a proper link 
* the link doesn't point to the AMP documentation

update-cache/lib/Signature.js
```diff
@@ -0,0 +1,43 @@
...
+    this._privateKey = privateKey;
+    this._sig = new jsrsasign.Signature({alg: 'SHA256withRSA'});
+    this._sig.init(this._privateKey);
+  }
+
+  /**
+   * Generates a signature for the content
+   * @param {string} data the signed content, using base64 url encoding.
+   */
+  create(data) {
```
sebastianbenz: suggestion: rename to 'generate'

update-cache/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+  // Load the private key, generated as described on 
+  // https://developers.google.com/amp/cache/update-cache#rsa-keys
+  const privateKey = '...';
+
+  // Create an instance of the factory using the private key.
+  const updateCacheUrlFactory = createUpdateCacheUrlFactory(privateKey);
+  const timestamp = (Date.now() / 1000) | 0; // timestamp as a UNIX Epoch in seconds.
+
+  const cacheUpdateUrls = await updateCacheUrlFactory.fromOriginUrl('https://www.example.com');
```
sebastianbenz: maybe use a promise instead of await (or wrap it in an async function) - otherwise people will get confused who are not familiar with async/await

update-cache/spec/UpdateCacheUrlFactorySpec.js
```diff
@@ -0,0 +1,46 @@
...
+
+describe('UpdateCacheUrl', () => {
+  const updateCacheUrl = new UpdateCacheUrlFactory(signature, caches);
+
+  describe('fromCacheUrl', () => {
+    it('Generates the correct signature', () => {
+      const timestamp = 1;
+      const result = updateCacheUrl.fromCacheUrl('https://test.com', timestamp);
+      const expected =
+          'https://test.com/update-cache/?amp_action=flush&amp_ts=1&amp_url_signature=';
```
sebastianbenz: how does this test if the signature is correct if it's an empty string?

update-cache/index.js
```diff
@@ -0,0 +1,38 @@
...
+const Caches = require('amp-toolbox-cache-list');
+
+/**
+ * Creates an instance of UpdateCacheUrlFactory that uses the privateKey
+ * to sign the Urls.
+ *
+ * @param {string} privateKey Private Key to be used when signing Urls.
+ * @returns {UpdateCacheUrlFactory} an instance of UpdateCacheUrlFactory.
+ */
+function createUpdateCacheUrlFactory(privateKey) {
```
sebastianbenz: Can we find a better name? A create method that creates a factory (which itself usually has a create method sounds wrong). 

How about 'createCacheUpdater(...)' and rename UpdateCacheUrlFactory to CacheUpdater?

andreban: The purpose of this method is simply to make the API simpler to use.

My first idea was creating a static `createFactory` on `UpdateCacheUrlFactory`, but I'd end up introducing a dependency for `Caches` and `Signature` to it. So, I decided moving this to `index.js`.

I've seem a few instance of Factories of Factories or a mix of Factory / Builders. Maybe it would make sense to rename `createFactory` to `buildFactory`?

`CacheUpdater` is not a good name - this class does not update the cache. It only creates the URLs that do it.

sebastianbenz: Hm, I'd prefer a static method in this case to avoid the double factory pattern. I really don't like the name Factory here, because it usually means a different thing (something that creates objects). You're right about CacheUpdater not being a good name though... 

How about:

```
class CacheUpdateUrlProvider {
 
  static create(privateKey) {
     const signature = new Signature(privateKey);
     const caches = new Caches();
     return new CacheUpdateUrlProvider(signature, caches);
  }

  constructor(..) {
  ...
}
```

I don't think it's a problem to have a dependency to Caches and Signature here, you have it anyway and if it is only at runtime.

update-cache/README.md
andreban: Couldn't find the update-cache spec o the AMP doc. I think it's specific to the Google AMP Cache. 

The AMP Cache guidelines state:

"8. Supports a public Update Cache mechanism for document publishers to notify the AMP cache about new, updated or deleted documents:

Equivalent to the Google AMP Update Cache API"

https://github.com/ampproject/amphtml/blob/00679960b7fd8e68f2ff159f84fe01bc8ea00a0f/spec/amp-cache-guidelines.md

update-cache/index.js
andreban: done

update-cache/spec/UpdateCacheUrlFactorySpec.js
andreban: done

update-cache/README.md
andreban: done.

update-cache/lib/Signature.js
andreban: done

update-cache/README.md
andreban: Changed to "Google AMP Cache documentation" instead of "AMP documentation"

andreban:
re: fromOriginUrl / calculateCacheUrls:

invalidation URLs can be created from the origin URL (eg: https://example.com) or from the cache URL (eg: https://example_com.cdn.ampproject.org/example.com/). 

The 2 methods,`fromOriginUrl` and `fromCacheUrl` reflect this. Would `calculateFromOriginUrl` / `calculateFromCacheUrl` work better?

sebastianbenz:
sgtm

andreban:
Addressed the comments. PTAL.

sebastianbenz:
One thing: we should update the readme to fully explain how to use the API, e.g. how to obtain a private key.

andreban:
Yes, I'll update the README

PR #53, opened 2018-06-11
Fix generating default timestamp

andreban:
- Generates default timestamp.
- Updates version to 0.1.1

### DIFF
--- a/update-cache/lib/UpdateCacheUrlProvider.js
+++ b/update-cache/lib/UpdateCacheUrlProvider.js
@@ -36,7 +36,7 @@ class UpdateCacheUrlProvider {
    * Generates update-cache URLs for each known AMP cache.
    *
    * @param {string} originUrl the URL for the content on the origin (ex: https://example.com)
-   * @param {Number} timestamp as a UNIX Epoch in seconds
+   * @param {Number} [timestamp] as a UNIX Epoch in seconds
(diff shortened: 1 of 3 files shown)


update-cache/lib/UpdateCacheUrlProvider.js
```diff
@@ -40,7 +40,7 @@ class UpdateCacheUrlProvider {
    * @returns {Array<Object>} an array with objects containing the cache ID, cache name and
    * update-cache url.
    */
-  calculateFromOriginUrl(originUrl, timestamp) {
+  calculateFromOriginUrl(originUrl, timestamp = defaultTimestamp_()) {
```
sebastianbenz: Can you also update the JS DOC to mark it as optional.

update-cache/lib/UpdateCacheUrlProvider.js
```diff
@@ -60,7 +60,7 @@ class UpdateCacheUrlProvider {
    * @param {Number} timestamp as a UNIX Epoch in seconds
```
sebastianbenz: Mark as optional

andreban: done.

update-cache/lib/UpdateCacheUrlProvider.js
andreban: done.

PR #54, opened 2018-06-11
Fix formatting and update doc

sebastianbenz:


sebastianbenz:
Should we also show a sample how to perform the actual request?

andreban:
Yes, I think it makes sense. 

Invoking `fetch(cacheUpdateUrlInfo.updateCacheUrl)` should be enough. We might want to parse the response and show results, but risking missing focus on the point of the sample, which is generating the URL.

PR #56, opened 2018-06-14
Move Resource Hints to top

andreban:
- Update OptimizeHeaderTransformer to move existing `preload`,
  `preconnect`, `prefetch` and `dns-prefetch` to top of page.
  Since those hints are not render-blocking, they are just below
  the charset meta tag.
- Keeps `prerender` hints at the end of the head, as the process
  is resource heavy and benefit the next click, not the current
  page being rendered.

Change-Id: I9d104507bb561b7d23db6ec7b0da0b61bd1b2b03

### DIFF
--- a/optimizer/lib/transformers/ReorderHeadTransformer.js
+++ b/optimizer/lib/transformers/ReorderHeadTransformer.js
@@ -27,6 +27,7 @@ class HeadNodes {
     this._metaOther = [];
     this._scriptRenderDelayingExtensions = [];
     this._scriptNonRenderDelayingExtensions = [];
+    this._resourceHintLinks = [];
     this._linkIcons = [];
     this._styleAmpCustom = null;
     this._linkStylesheetsBeforeAmpCustom = [];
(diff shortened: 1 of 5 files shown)


optimizer/lib/transformers/ReorderHeadTransformer.js
```diff
@@ -157,6 +159,15 @@ class HeadNodes {
       this._linkIcons.push(node);
       return;
     }
+
+    if (rel === 'preload' ||
+      rel === 'prefetch' ||
+      rel === 'dns-prefetch' ||
+      rel === 'preconnect') {
```
sebastianbenz: nit: I'd prefer the different values in a constant HashMap. Feel free to ignore though...

andreban: This is in my upper bound of tolerance too. But I opted to keep the same style as the block above. If we add more, then we definitely need to change to improve readability.

PR #66, opened 2018-06-21
add preloads after viewport

### LINKED ISSUE #63: Place preloads after viewport
preloads need to be declared after 

```
meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
```

otherwise media queries cannot be evaluated (when combined with preload)

sebastianbenz:
Fixes #63

### DIFF
--- a/cache-list/package-lock.json
+++ b/cache-list/package-lock.json
@@ -16,7 +16,7 @@
       "integrity": "sha1-r9+UiPsezvyDSPb7IvRk4ypYs2s=",
       "dev": true,
       "requires": {
-        "acorn": "3.3.0"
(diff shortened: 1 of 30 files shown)


andreban:
We are assuming `meta name=viewport` will always come after `meta charset`, right? 

I think that's reasonable, otherwise the original AMP would be invalid anyway.

The other options would be to find viewport and charset, and insert after the one that shows up later in the code.

sebastianbenz:
meta charset can be anywhere in the head and the file will still be valid. Hence, I think the correct approach is to add it after viewport as it's the only hard constraint we have.

PR #67, opened 2018-06-21
PreloadImages should ignore templates

### LINKED ISSUE #64: Preload image transformer should ignore images in amp-list/mustache template
Image preload can result in `<link rel="preload" href="{{image}}?w311h205" as="image">`
When there is an amp-list mustache template e.g.

```
<amp-list src="https://www.standard.co.uk/pwamp/layout_component/sidebar-videos" items="articles" class="amp-list list-videos" layout="fixed-height" width="auto" height="370" media="(min-width: 930px)">
    <template type="amp-mustache">
        <a class="amp-list-item" href="{{link}}" data-vars-event-id="c23" data-vars-item-name="Popular Videos - {{link}}">
            <amp-img src="{{image}}?w311h205" width="96" height="64" layout="responsive" class="list-item-img"></amp-img>
            <span class="list-item-text">{{title}}</span>
        </a>
    </template>
</amp-list>
```

sebastianbenz:
Fixes #64

### DIFF
--- a/optimizer/lib/transformers/PreloadImages.js
+++ b/optimizer/lib/transformers/PreloadImages.js
@@ -39,23 +39,18 @@ class PreloadImages {
     const body = html.firstChildByTag('body');
     const preloadImageMap = new Map();
 
-    for (let node = body; node !== null; node = node.nextNode()) {
+    let node = body;
+    while (node !== null) {
       // We've hit the maximum number of preloads.
       if (preloadImageMap.size >= MAX_PRELOADED_IMAGES) {
         break;
       }
-      let imageUrl = this.extractImageUrl(node);
-
-      if (!imageUrl) {
-        continue;
+      if (node.tagName === 'template') {
+        node = node.nextSibling;
+      } else {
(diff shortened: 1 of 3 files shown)


optimizer/lib/transformers/PreloadImages.js
```diff
@@ -39,23 +39,18 @@ class PreloadImages {
...
-      let imageUrl = this.extractImageUrl(node);
-
-      if (!imageUrl) {
-        continue;
-      }
-
-      // If srcset is used, skip preloading as we don't know which image will be used.
-      if (node.attribs.srcset) {
-        continue;
+      if (node.tagName === 'template') {
```
andreban: I think this may break with something like this:
`
<html>
  <div>
    <h1></h1>
    <template></template>
  </div>
  <amp-img src='...'/>
</html>
`
When it gets to `template`, `node.nextSibling` will be `null`, which will then exit the loop and not process the image below it.

sebastianbenz: done

PR #70, opened 2018-06-26
Fixes runtime-css replacing > with &gt;

andreban:
- Favors parse5 insertText method instead of using our own
  method to create a and insert a text node.

### DIFF
--- a/optimizer/lib/TreeParser.js
+++ b/optimizer/lib/TreeParser.js
@@ -113,6 +113,10 @@ Node.hasAttribute = function(attribute) {
   return attribute in this.attribs;
 };
 
+Node.insertText = function(text) {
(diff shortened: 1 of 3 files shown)


optimizer/lib/TreeParser.js
```diff
@@ -113,6 +113,10 @@ Node.hasAttribute = function(attribute) {
   return attribute in this.attribs;
 };
 
+Node.insertText = function(text) {
+  htmlparser2.insertText(this, text);
```
ithinkihaveacat: Should this return something?

andreban: The original [insertText](https://github.com/ampproject/amp-toolbox/pull/70) doesn't return anything.

Issue #62, opened 2018-06-18
Handle existing preloads


sebastianbenz:
* Avoid duplicating preloads (e.g. Google fonts)
* rewrite existing v0.js preload if one already exists (for versioned URLs)

PR #73, opened 2018-07-20
Pull request to prune duplicate preload links from the header.

dakkad:
#62 Solution which only looks for preload duplicate href in headers.


optimizer/lib/transformers/PruneDuplicatePreloads.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+    // replace the child node list
+    head.childNodes = childNodes;
+  }
+
+  _notPreloadLink(node) {
+    if (node.tagName !== 'link') {
+      return true;
+    }
+    if (!node.attribs || !node.attribs.rel) {
```
sebastianbenz: nit: ` || !node.attribs.rel` is not needed here

optimizer/lib/transformers/PruneDuplicatePreloads.js
```diff
@@ -0,0 +1,78 @@
...
+      return true;
+    }
+    return node.attribs.rel !== 'preload';
+  }
+
+  _alreadyLoaded(link, preloaded) {
+    if (!link.attribs || !link.attribs.href) {
```
sebastianbenz: nit: this check is not needed here as we already know it's a valid preload link.

sebastianbenz: to make this more obvious you could rename `link` to `linkPreload`

optimizer/lib/transformers/PruneDuplicatePreloads.js
```diff
@@ -0,0 +1,78 @@
...
+      return false;
+    }
+    return preloaded.has(link.attribs.href);
+  }
+
+  _markPreloaded(link, preloaded) {
+    if (!link.attribs || !link.attribs.href) {
```
sebastianbenz: nit: same as above, the check is not strictly necessary.

optimizer/lib/transformers/PruneDuplicatePreloads.js
```diff
@@ -0,0 +1,85 @@
...
+    return relations.has(rel);
+  }
+
+  _markPreloaded(link, preloaded) {
+    let rel = link.attribs.rel;
+    let href = link.attribs.href;
+    if (!preloaded.has(href)) {
+      preloaded.set(href, new Set());
+    }
+    let relations = preloaded.get(href);
```
sebastianbenz: Suggestion: rewrite to avoid reading from the map twice:

```
let relations = preloaded.get(href)
if (!relations) {
  relations = new Set();
  preloaded.set(href, relations);
}
```

optimizer/lib/transformers/PruneDuplicatePreloads.js
```diff
@@ -0,0 +1,85 @@
...
+    return relations.has(rel);
+  }
+
+  _markPreloaded(link, preloaded) {
+    let rel = link.attribs.rel;
+    let href = link.attribs.href;
+    if (!preloaded.has(href)) {
+      preloaded.set(href, new Set());
```
sebastianbenz: A simpler approach would be to concatenate relevant and href as key:

```
preloaded.set(this.keyFrom(link), link);

if (preloaded.has(this.keyFrom(link))) {
   ...
}

... 

keyFrom(link) {
  return link.attribs.rel + link.attribs.href;
}
```

optimizer/lib/transformers/PruneDuplicatePreloads.js
```diff
@@ -0,0 +1,85 @@
...
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+/**
+ * Removes duplicate preload header directives.
```
sebastianbenz: s/preload/resource hint

optimizer/index.js
```diff
@@ -33,7 +33,8 @@ const defaultConfig = {
     'ReorderHeadTransformer',
     // needs to run after ReorderHeadTransformer
     'RewriteAmpUrls',
-    'GoogleFontsPreconnect'
+    'GoogleFontsPreconnect',
+    'PruneDuplicatePreloads'
```
sebastianbenz: suggestion: rename to `PruneDuplicateResourceHints`

optimizer/lib/transformers/PruneDuplicatePreloads.js
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * To avoid wasted requests for preloaded resources strip references to duplicate
+ * items.
+ */
+class PruneDuplicatePreloads {
+
+  transform(tree) {
+    const preloaded = new Map();
```
sebastianbenz: s/preloaded/resourceHints

optimizer/lib/transformers/PruneDuplicatePreloads.js
dakkad: Done.

optimizer/lib/transformers/PruneDuplicatePreloads.js
dakkad: I agree that it is simpler. I didn't do it that way as it seems more likely to fail in situations where the rel / href clash etc. e.g. rel = '' and href = 'preload' matches rel='preload' and href = ''. I can simplify if you are happy with this behaviour?

optimizer/index.js
dakkad: Done.

optimizer/lib/transformers/PruneDuplicatePreloads.js
dakkad: Done.

optimizer/lib/transformers/PruneDuplicatePreloads.js
sebastianbenz: It should only be a problem for resource hints that are broken anyway...and in that case a clash is probably a good thing. But I'm fine with the way it is.

sebastianbenz:
There's one problem we forgot to address: what about cases where preloads have different priorities, e.g.:

```
 <link rel="preload" href="https://cdn.ampproject.org/v0.js">
 <link rel="preload" as="script" href="https://cdn.ampproject.org/v0.js">
```

In that case, only the more precise `as=script` resource should be kept.

dakkad:
Interesting point RE:

```
 <link rel="preload" href="https://cdn.ampproject.org/v0.js">
 <link rel="preload" as="script" href="https://cdn.ampproject.org/v0.js">
```

In the current implementation you are correct that the second (valid) link would be pruned. How about solving for 'invalid' input in an earlier transformer? 

I propose adding this check (that there is 'as') as a prequalifier for a link that can be pruned so in the example above the first link would not be impacted?

PR #97, opened 2018-09-27
remove no longer valid test for version prefix

sebastianbenz:


sebastianbenz:
Most of the time is not good for testing ;-).

choumx:
Agreed, just FYI. :)

PR #99, opened 2018-10-02
Insert hidden attribute when layout=nodisplay

jridgewell:
Context: https://github.com/ampproject/amphtml/pull/18419

AMP Runtime now controls display:none toggling through the hidden attribute. This updates the SSR transform to inject the hidden attribute instead of an inline display:none style.

This must be merged **after** the next production promote. I'll comment when that happens.


sebastianbenz:
Is this live?

jridgewell:
The next promotion happens sometime today. We can merge just after that.

jridgewell:
@sebastianbenz: Ready to merge.

sebastianbenz:
Thanks!

PR #105, opened 2018-10-05
Cache URL: Clean Up for 1.0.0, Examples, and Improved bundle sizes

torch2424:
Forewarning: PR size looks large due to the `package-lock.json`.

This does a final sanity check, and clean up for the module. Completed the following tasks:

* Moved the module to ES6 Syntax (better tree shaking this way)
* Added a rollupplugin to show our bundle sizes for each platform
* Created an `examples` directory, that is hosted with `npm run serve`, and can be used for manual testing of the module ES6 output
* Introduced @kristoferbaxter AMP closure compiler plugin 👍 
* Used the CommonJs output in tests, as opposed to the unbuilt lib
* Removed the need of node builtins in browser module output, by using [url-parse](https://www.npmjs.com/package/url-parse) for a cross platform string -> URL object solution

cc @erwinmombay , because I made a few more rollup config changes, want to make sure it's still looking okay on their end.

# Screenshots

screen shot 2018-10-05 at 12 07 13 pm

screen shot 2018-10-05 at 4 01 47 pm


torch2424:
Thank you!

Fixed the conflicts, and this is now ready to be pulled. If you get the chance, would you be so kind to deploy to npm? That way I can start using the package on other AMP projects.

Thanks again for working on this with me! 😄

torch2424:
Made the windows fixes, this is good to go! 😄 

Just in case my other comment is scrolled by:

> If you get the chance after merging this, could you deploy to npm? That way I can start using the package on other AMP projects. Thanks again for working on this with me! 😄

sebastianbenz:
It's published. Thanks Aaron!

torch2424:
Thank you so much @sebastianbenz ! Glad we could get this updated 😀

PR #113, opened 2018-10-10
amp-toolbox-cache-url: Specified files in package.json

torch2424:
Noticed on unpkg, which is a service that acts as a mirror and CDN of npm, we weren't uploading built files to the module: 

https://unpkg.com/amp-toolbox-cache-url@1.0.0/

This includes the [files key to our package.json](https://docs.npmjs.com/files/package.json#files), that way we include the build output.

**Please note:** This required that we build the module before we publish to npm. Perhaps we should make a deploy script @sebastianbenz ?

### DIFF
--- a/cache-url/package.json
+++ b/cache-url/package.json
@@ -1,6 +1,6 @@
 {
   "name": "amp-toolbox-cache-url",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "Transform canonical URLs into AMP Cache URLs",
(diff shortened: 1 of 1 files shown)


torch2424:
Also, my apologies on missing this! This shouldn't be a huge problem, as anyone using this exact version shouldn't incur any breaking changes. But it is mostly helpful for not having to run babel on the module.

After this gets merged, let's go ahead and deploy this out 😄

sebastianbenz:
...and it's published

PR #115, opened 2018-10-10
Update Cache API 1.0

sebastianbenz:
* uses Cache URL 1.0
* migrates API to promises as cache URL is now promise based
* adds prepublish script for cache URL package

### DIFF
--- a/cache-url/package.json
+++ b/cache-url/package.json
@@ -1,12 +1,13 @@
 {
   "name": "amp-toolbox-cache-url",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "description": "Transform canonical URLs into AMP Cache URLs",
   "main": "dist/amp-toolbox-cache-url.cjs.js",
   "module": "dist/amp-toolbox-cache-url.esm.js",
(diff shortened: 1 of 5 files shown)


cache-url/package.json
```diff
@@ -1,12 +1,13 @@
...
   "name": "amp-toolbox-cache-url",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "description": "Transform canonical URLs into AMP Cache URLs",
   "main": "dist/amp-toolbox-cache-url.cjs.js",
   "module": "dist/amp-toolbox-cache-url.esm.js",
   "browser": "dist/amp-toolbox-cache-url.umd.js",
   "scripts": {
     "build": "npx rollup -c",
+    "prepublish": "npm run build",
```
sebastianbenz: @torch2424 added a prepublish script to ensure that build is run before publishing to NPM (I borked this when publishing 1.0.1).

torch2424: No worries! I shouldn't have forgotten the "files" key to begin with haha! 😄 I'm just so used to it already being there I forgot. Thank you very much for this!

PR #128, opened 2018-10-19
amp-toolbox-cache-url: Used Rollup Plugin Replace for Platform Specific code

torch2424:
# Context

relates to: https://github.com/ampproject/amphtml/pull/18676
cc @erwinmombay 

PR is small, `package-lock.json` changes cause inflation.

# Description

After working on the PR mentioned above, we found that the closure compiler we are using the in the main AMP project doesn't allow require statements, even if guarded by an `if` block, and with warnings suppressed. And after implementing the exact same change here in a personal project, realized this can also be used here to accomplish the same task. This also has the benefit of not shipping down code that we will not use in the browser 😄 

At build-time, rollup will remove the specified comments for the correct platform, thus enabling the code to be the true function of `_sha256`. If on browser, our `@ampproject/closure-compiler` will go ahead and minify / remove the `ROLLUP_REPLACE_NODE` comments as well 👍 

# Next Steps

Deploy this on npm for version `1.0.3`. Version is already bumped in this PR 😄

### DIFF
--- a/cache-url/lib/AmpCurlUrlGenerator.js
+++ b/cache-url/lib/AmpCurlUrlGenerator.js
@@ -132,25 +132,30 @@ function constructFallbackCurlsCacheDomain_(domain) {
 
 
 /**
+ * Function to run sha256 over a string.
+ * Uses rollup-plugin-replace, to import/use any platform
+ * specific crypto code, and then use accordingly.
  * @param {string} str The string to convert to sha256
  * @return {!Promise<string>}
  * @private
  */
 function sha256_(str) {
-  if (typeof window !== 'undefined') {
-    // Transform the string into an arraybuffer.
-    const buffer = new TextEncoder('utf-8').encode(str);
-    return crypto.subtle.digest('SHA-256', buffer).then((hash) => {
-      return hex_(hash);
-    });
-  } else {
-    const buffer = Buffer.from(str, 'utf-8');
-    const crypto = require('crypto');
-    return new Promise((resolve) => {
-      const sha256 = crypto.createHash('sha256').update(buffer).digest('hex');
-      resolve(sha256);
-    });
-  }
+  /*ROLLUP_REPLACE_BROWSER
+  // Transform the string into an arraybuffer.
+  const buffer = new TextEncoder('utf-8').encode(str);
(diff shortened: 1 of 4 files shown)


torch2424:
cc @sebastianbenz This is ready for review whenever you get the time 😄 No rush, as we have a workaround in the current `ampproject/amphtml` for now. And will move this over once merged 👍

cache-url/lib/AmpCurlUrlGenerator.js
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+/* eslint no-unused-vars: ["error", { "varsIgnorePattern": "hex_", "argsIgnorePattern": "str" }]*/
```
sebastianbenz: Do we need to ignore `hex_`?

sebastianbenz: Do we need to ignore `hex_`?

sebastianbenz: Do we need to ignore `hex_`?

sebastianbenz: Do we need to ignore `hex_`?

torch2424: Yes, `hex_` is used by SHA256, therefore, since we are commenting it, it is considered as unused 😢

torch2424:
@sebastianbenz No worries! Yeah Github has been having data issues for a bit, it's on their status page.

I replied to the `hex_` comment, TL;DR yes.

This is the only way I can think of personally, but @kristoferbaxter may know another solution. Would like to hear their opinion 😄

torch2424:
Talked to @kristoferbaxter offline, and found that we could do the following:

Abstract our "platform specific code" into two separate exports in separate files, and use: https://github.com/ember-cli/babel-plugin-filter-imports to filter the imports per build platform. That way we get all the benefits of linting and everything 😄 

@sebastianbenz You okay with going with this solution instead?

sebastianbenz:
That sounds great! Thanks a lot for looking into this!

torch2424:
Closing in favor of #132

PR #131, opened 2018-10-23
update eslint-config-google to 0.11.0

sebastianbenz:


sebastianbenz:
@ithinkihaveacat PTAL as Andre is OOO

sebastianbenz:
1. Yes, this seems to be a lerna best practices. Added it to gitignore.
2. Made things worth (and better). All tests are moved to the root folder. Removes lots of duplication and speeds things up. It becomes harder though to run package specific tests.
3. Yes.

.travis.yml
```diff
@@ -6,5 +6,9 @@ node_js:
 sudo: required
 addons:
   chrome: stable 
-install: npm install -g npx && npx lerna bootstrap  
+cache:
+  directories:
+    - ~/.npm
+before_install: if [[ `npm -v` != 6* ]]; then npm i -g npm@6; fi
```
ithinkihaveacat: Maybe should document that npm 6 is required for local install as well. Might be possible to use the "prepare" script or something to actually test for this instead.  FFS `npm`…

package.json
```diff
@@ -1,7 +1,48 @@
 {
   "name": "root",
   "private": true,
+  "scripts": {
+    "build": "lerna run build",
```
ithinkihaveacat: Should this be `npx lerna run build`? It's not otherwise installed.

sebastianbenz: done

.travis.yml
sebastianbenz: I plan to document the dev setup in another PR.

sebastianbenz:
Thanks Michael! Should all be fixed now. PTAL

PR #149, opened 2018-11-12
Improve Debugging & Configuation

sebastianbenz:
This commit does several things:

1. Introduces a simple core logger supporting colored terminal output.
2. Migrated CLI logger to core logger.
3. Add option for verbose logging output.
4. Introduce two configs (SRR and valid AMPs), which are not yet
publicly documented.


sebastianbenz:
Log output looks like this:

screenshot 2018-11-12 at 15 59 35

packages/optimizer/lib/DomTransformer.js
```diff
@@ -49,28 +93,46 @@ class DomTransformer {
...
       });
     };
-    return this._transformers.reduce(sequence, Promise.resolve());
+    return this.transformers_.reduce(sequence, Promise.resolve());
   }
 
   /**
    * Set the config.
    * @param {Object} config - The config.
+   * @param {boolean} config.verbose - true if verbose mode should be enabled.
```
andreban: Seems like we are missing the `config.validAmp` parameter here.

packages/optimizer/lib/DomTransformer.js
```diff
@@ -17,6 +17,50 @@
...
+  // needs to run after ReorderHeadTransformer
+  'RewriteAmpUrls',
+  'GoogleFontsPreconnect',
+  'PruneDuplicateResourceHints',
+  'AddBlurryImagePlaceholders',
+];
+
+const DEFAULT_CONFIG = {
+  verbose: false,
+  validAmp: false,
```
andreban: It seems that both `validAmp` and `transformers` control which transformers will run? I'm wondering if it be more clear to provide factory methods to create the transformers. eg:

`DomTransformer.allTransformations(verbose: false)` and `DomTransformar.validAmp(verbose: false)`

Those would create a config object and call the constructor with the value for verbose and the correct list of transformers.

sebastianbenz: I'm thinking that it probably would be best to get away from users having to handle transformers on their own. I think it'd be better to handle everything via configuration:

```
const optimiser = AmpOptimizer.create({
  validAmp: true,
  verboseTrue: true,
  preProcessors: [ ... list of transformers ... ],
  postProcessors: [ ... list of transformers ... ]
});
```

sebastianbenz: This PR works towards this, but doesn't break existing APIs yet.

andreban: We're going to have more PRs moving in that direction then. Sounds Good.

packages/optimizer/lib/DomTransformer.js
sebastianbenz: done

sebastianbenz:
PTAL

andreban:
LGTM, Thanks!

sebastianbenz:
thanks for the review

PR #150, opened 2018-11-13
Add AMP CORS middleware

sebastianbenz:


sebastianbenz:
Not sure about how to best approach origin checking. Which caches should be whitelisted by default, cdn.ampproject.org/caches + bing?

packages/cors/lib/cors.js
```diff
@@ -0,0 +1,81 @@
...
+      return;
+    }
+    // Check if sourceOrigin is allowed
+    if (options.sourceOriginPattern && !options.sourceOriginPattern.test(sourceOrigin)) {
+      response.status(403).end(); // forbidden
+      return;
+    }
+    // Get either the source origin or the cache origin
+    const origin = calculateOrigin_(request.headers, sourceOrigin);
+    if (!origin) {
```
sebastianbenz: @andreban wondering if we should return `400` if __amp_source_origin, but neither `Origin` nor `AMP-SAME-ORIGIN`?

andreban: > Not sure about how to best approach origin checking. Which caches should be whitelisted by default, cdn.ampproject.org/caches + bing?

Yes, I think that makes sense.

andreban: > @andreban wondering if we should return `400` if __amp_source_origin, but neither `Origin` nor `AMP-SAME-ORIGIN`?

I think both are OK? You could see it as missing information that could generate a Bad Request, or information that doesn't give access to the content that would give Access Forbidden.

sebastianbenz: returning a 400 now

sebastianbenz:
Origins are now filtered (bing is hardcoded). Also added logging.

PTAL

sebastianbenz:
OK, made a few last changes, but that should be it (for now...).

sebastianbenz:
OK, small improvements. Test failures are caused by #156.

packages/cors/README.md
```diff
@@ -0,0 +1,108 @@
...
+  sourceOriginPattern: /https:\/\/ampbyexample\.com$/
+}));
+```
+
+This will only allow requests with `https://ampbyexample.com` set as the source origin. Requests from all other origins
+will receive a `403` response,
+
+### Origin verification
+
+By default, the AMP CPRS middleware will only allow requests from AMP Caches listed on
```
andreban: s/CPRS/CORS

PR #152, opened 2018-11-14
cache-url: Replaced the Text Encoder with a More Browser Compatible Solution

### LINKED ISSUE #135: cache-url: Remove TextEncoder
see #132 

TextEncoder is not supported on some browsers, thus it would be good to remove this and use something else.

cc @sebastianbenz @kristoferbaxter

torch2424:
closes #135 

This replaces the Text Encode with a solution I made at: https://jsfiddle.net/torch2424/nhma9q3p

# Example

Added a simple log to make sure we were doing the sha256, not in PR. But shows we are getting the exact same hashes, and are producing the same buffers for our use case 😄 

screen shot 2018-11-13 at 5 43 21 pm

### DIFF
--- a/packages/cache-url/lib/browser/Sha256.js
+++ b/packages/cache-url/lib/browser/Sha256.js
@@ -21,14 +21,28 @@
  */
 export default function browserSha256(str) {
   // Transform the string into an arraybuffer.
-  const buffer = new TextEncoder('utf-8').encode(str);
+  const buffer = stringToBuffer_(str);
   return crypto.subtle.digest('SHA-256', buffer).then((hash) => {
     return hex_(hash);
   });
 }
 
 /**
- * @param {string} buffer
+ * Function to convert a UTF-8 String to a buffer
+ * @param {string} str
+ * @return {Array<number>}
+ * @private
+ */
+function stringToBuffer_(str) {
+  const array = new Uint8Array(str.length);
+  for (let i = 0; i < str.length; i++) {
+    array[i] = str.charCodeAt(i);
+  }
+  return array.buffer;
+}
+
+/**
+ * @param {Array<number>} buffer
  * @return {string}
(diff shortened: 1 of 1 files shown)


torch2424:
cc @sebastianbenz

packages/cache-url/lib/browser/Sha256.js
```diff
@@ -21,14 +21,28 @@
...
+ * @param {string} str
+ * @return {Array<number>}
+ * @private
+ */
+function stringToBuffer_(str) {
+  const array = new Uint8Array(str.length);
+  for (let i = 0; i < str.length; i++) {
+    array[i] = str.charCodeAt(i);
+  }
+  return array.buffer;
```
sebastianbenz: Doesn't TextEncoder#encode return an Uint8Array?

torch2424: Oh huh, you are correct: https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/encode

Odd that it still works...

I'll go ahead and make the change 😄

torch2424:
@sebastianbenz Ah, thanks for testing this!

In the test, we already have a few i18n domains in the tests, see: 

https://github.com/ampproject/amp-toolbox/blob/master/packages/cache-url/spec/AmpUrlSpec.js#L119

But this is something I'll look into 😄

torch2424:
Also, I just opened up #154 I'd like to get this figured out before I finish this up, just because I had to hack around to get the tests to run for this. And, I think it would be better ot just fix it first haha! 😄

sebastianbenz:
Closing this until we find a better way.

torch2424:
@sebastianbenz Thanks for updating this. Still on my todo list, and will try to fit this into our next fixit 😄

PR #227, opened 2019-02-18
Use default amp=1 prefix

sebastianbenz:
Instead of `amp=`


sebastianbenz:
Thanks for the review. This shouldn't break anyone as we only [check for the existence of the query parameter](https://github.com/ampproject/amp-toolbox/blob/master/packages/optimizer-express/lib/UrlMapping.js#L39). This is purely cosmetics.

PR #233, opened 2019-02-21
fallback to v0css link 

sebastianbenz:


sebastianbenz:
Good question. It's a pathological case (which makes it possible to run the optimiser while offline). Usually, the version wouldn't be set if there's no internet connection.

PR #249, opened 2019-03-12
Placeholder blurry disable

sebastianbenz:


sebastianbenz:
@andreban what do you think of this? It's a breaking change, but I think we should make it.

packages/optimizer/README.md
```diff
@@ -83,7 +83,8 @@ You can find a sample implementation [here](demo/simple/). If you're using the e
 Currently the following options are supported:
 
 * **ampUrl**: an URL string pointing to the valid AMP version. Required by the AddAmpLink transformer.
-* **imageBasePath**: a base URL or path required for resolving an image file from a given image `src` attribute. Required by the AddBlurryImagePlaceholder transformer.
+* **blurredPlaceholders**: Enables blurry image placeholder generation. Default is `false`. Make sure to only enable  **Important:** blurry image placeholder computation is expensive. Make sure to only use it for static or cached pages.
```
andreban: This sentence looks incomplete:  "make sure to only enable"

andreban: Also, for consistency: s/Enable/enable

andreban:
+1. I think folks using the blurredPlaceholder should make a conscious decision.

packages/optimizer/README.md

sebastianbenz:
@andreban PTAL

PR #260, opened 2019-03-25
Placeholder cache

sebastianbenz:
This adds an LRU cache for placeholder calculation. This also introduces the need for a new config level (when the optimizer gets created). We will have to move many of the options to this config level in the future. 

At the moment, it's not possible to completely disable the cache. Not sure if I should add a flag to disable it.

### DIFF
--- a/packages/optimizer/lib/transformers/AddBlurryImagePlaceholders.js
+++ b/packages/optimizer/lib/transformers/AddBlurryImagePlaceholders.js
@@ -23,7 +23,9 @@ const PathResolver = require('../PathResolver');
 const log = require('../log').tag('AddBlurryImagePlaceholders');
 
 const PIXEL_TARGET = 60;
-const MAX_BLURRED_PLACEHOLDERS = 5;
+const MAX_BLURRED_PLACEHOLDERS = 100;
+const DEFAULT_CACHED_PLACEHOLDERS = 30;
+
 const ESCAPE_TABLE = {
   '#': '%23',
   '%': '%25',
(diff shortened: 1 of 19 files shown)


packages/optimizer/lib/transformers/AddBlurryImagePlaceholders.js
```diff
@@ -48,11 +52,28 @@ function escaper(match) {
...
  * * `maxBlurredPlaceholders`: Specifies the max number of blurred images. Defaults to 5.
+ * * `blurredPlaceholdersCacheSize`: Specifies the max number of blurred images to be cached
+ *   to avoid expensive recalculation. Set to 0 if all placeholders should be cached. Defaults
+ *   to 30.
  *
  * Important: blurry image placeholder computation is expensive. Make sure to
  * only use it for static or cached pages.
  */
 class AddBlurryImagePlaceholders {
+  constructor(config) {
```
andreban: nit: The next two lines could be written as:

```
constructor({blurredPlaceholdersCacheSize = DEFAULT_CACHED_PLACEHOLDERS} = {}) {
```

packages/optimizer/lib/transformers/AmpBoilerplateTransformer.js
```diff
@@ -28,7 +28,7 @@ const V0_CSS_URL = AMP_CACHE_HOST + '/' + V0_CSS;
  * runtime version is specified, v0.css will be inlined.
  */
 class AmpBoilerplateTransformer {
-  constructor(fetch = OneBehindFetch.create()) {
+  constructor(config, fetch = OneBehindFetch.create()) {
```
andreban: nit: I'm wondering if fetch should be a parameter in itself or another option in the config.

sebastianbenz: Yes, that could make sense. We should keep it in mind.

packages/optimizer/lib/transformers/AddBlurryImagePlaceholders.js
sebastianbenz: that's too much magic for me

PR #267, opened 2019-04-03
Add check to SeparateKeyframes to only move valid keyframes

ijjk:
After implementing the optimizer into Next.js I ran into pages with keyframes that use properties other than the allowed few would no longer validate and the validator gives the below message.  

`error  CSS syntax error in tag 'style[amp-keyframes]' - invalid property 'stroke-dashoffset'. The only allowed properties are '['animation-timing-function', 'offset-distance', 'opacity', 'transform', 'visibility']`

This just adds a check to make sure the keyframes are valid before moving them since copying the keyframes and only moving the valid properties would inflate the page size. Thanks for your help getting the initial PR shipped so quickly!


packages/optimizer/lib/transformers/SeparateKeyframes.js
```diff
@@ -17,6 +17,14 @@
 
 const css = require('css');
 const stringifyOptions = {indent: 0, compress: true};
+
+const allowedKeyframeProps = {
```
sebastianbenz: It's ok to use a [Set](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&cad=rja&uact=8&ved=2ahUKEwjSn_Pmn7PhAhVD0aYKHQdECP0QFjAAegQIAhAB&url=https%3A%2F%2Fdeveloper.mozilla.org%2Fde%2Fdocs%2FWeb%2FJavaScript%2FReference%2FGlobal_Objects%2FSet&usg=AOvVaw2eLvvECVNKoL7bdb7l6uhS) here.

packages/optimizer/lib/transformers/SeparateKeyframes.js
```diff
@@ -60,8 +68,19 @@ class SeparateKeyframes {
...
+          return !allowedKeyframeProps[d.property];
+        });
+      });
+    };
+
     cssTree.stylesheet.rules = cssTree.stylesheet.rules.filter((rule) => {
       if (rule.type === 'keyframes') {
+        // We can't move a keyframe with an invalid property
+        // or else the style[amp-keyframes] is invalid
+        if (isInvalidKeyframe(rule)) return true;
```
sebastianbenz: Can you log a warning here:

```
const log = require('../log').tag('SeparateKeyframes');
...

log.warn('...').
```

Ideally, with some context making it easy to identify the invalid keyframe.

packages/optimizer/spec/transformers/SeparateKeyframes/doesnt_move_invalid_keyframes/input.html
```diff
@@ -0,0 +1,36 @@
...
+<html ⚡>
+<head>
+  <script async custom-element=amp-experiment src=https://cdn.ampproject.org/v0/amp-experiment-0.1.js></script>
+  <script async src="https://cdn.ampproject.org/v0.js"></script>
+  <link rel=stylesheet href=https://cdn.ampproject.org/v0.css>
+  <link href=https://fonts.googleapis.com/css?foobar rel=stylesheet type=text/css>
+  <link href=https://example.com/favicon.ico rel=icon>
+  <style amp-custom>
+    .another{ display:block }
+    @keyframes anim1{
```
sebastianbenz: suggestion: name the keyframe in a way to identify to expected behavior, e.g. `@keyframes invalidHeight`.

packages/optimizer/lib/transformers/SeparateKeyframes.js
```diff
@@ -60,8 +68,19 @@ class SeparateKeyframes {
       },
     };
 
+    const isInvalidKeyframe = (keyframe) => {
+      return keyframe.keyframes.some((k) => {
```
sebastianbenz: A good old nested for loop might be more performant here.

packages/optimizer/lib/transformers/SeparateKeyframes.js
ijjk: Ok cool that's better, updated

packages/optimizer/lib/transformers/SeparateKeyframes.js
ijjk: Updated

packages/optimizer/spec/transformers/SeparateKeyframes/doesnt_move_invalid_keyframes/input.html
ijjk: Good idea, updated

packages/optimizer/lib/transformers/SeparateKeyframes.js
ijjk: Updated

PR #270, opened 2019-04-03
Make jimp a optional peerdependency

timneutkens:
The way this is implemented is to avoid peerDependency warnings. We saw this issue with `unistore` where users would blindly install `preact`/`react` the moment they saw a peerdependency warning, even though the package doesn't need it to function correctly.

I've also updated the documentation.

### DIFF
--- a/packages/optimizer/README.md
+++ b/packages/optimizer/README.md
@@ -83,7 +83,7 @@ You can find a sample implementation [here](demo/simple/). If you're using the e
 Currently the following options are supported:
 
 * **ampUrl**: an URL string pointing to the valid AMP version. Required by the AddAmpLink transformer.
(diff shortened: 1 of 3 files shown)


sebastianbenz:
Thanks Tim! The basic approach LGTM. However, we need to guard requiring jimp [here](https://github.com/ampproject/amp-toolbox/blob/master/packages/optimizer/lib/transformers/AddBlurryImagePlaceholders.js#L18) as the transformer will still be instantiated. 

We could then also print an error if jimp is not installed here:

```
    if (!params.blurredPlaceholders) {
      return;
    }
    if (!jimp) {
       log.error('jimp needs to be installed via `npm install jimp` for this transformer to work');
       return;
    }
```

sebastianbenz:
Can you add both as a devDependency to ensure that tests run successfully.

PR #275, opened 2019-04-05
Migrate axios to node-fetch

sebastianbenz:
Use node-fetch instead of axios for the smaller bundle size. This commit
also refactors OneBehindFetch to be an API compatible wrapper around the
fetch API.

One positive side-effect is that the fetch implementation now becomes an
option of AMP Optimizer.


sebastianbenz:
There's still some more work todo.

sebastianbenz:
@andreban OK - works now. PTAL

packages/core/lib/OneBehindFetch.js
```diff
@@ -15,58 +15,46 @@
...
-  /**
-   * Creates a new OneBehindFetch.
-   *
-   * @param {Axios} axio request handler
-   * @returns {OneBehindFetch}
-   */
-  constructor(axios) {
-    this.cache_ = {};
-    this.axios_ = axios;
+async function oneBehindFetch(input, init) {
```
andreban: This function is now with a mix of async / await and Promise syntax. I find this a bit confusing. Consider refactoring to settle on a single syntax. 

eg: One the last line, this could work: `return (await result).clone()`, but it may not be clear enough and breaking into something like the below may be clearer:

```
const response = await result;
return result.clone();
```

packages/cache-list/lib/AmpCaches.js
```diff
@@ -27,11 +27,10 @@ class AmpCaches {
   /**
    * Creates a new instance of AmpCaches.
    *
-   * @param {Object} [fetchStrategy = OneBehindFetch] The fetch strategy to be used when fetching
-   * data from the caches endpoint. Defaults to OneBehindFetch.
+   * @param {Function} fetch - a fetch implementation
    */
-  constructor(fetchStrategy = OneBehindFetch.create()) {
-    this.fetchStrategy_ = fetchStrategy;
+  constructor(fetch=oneBehindFetch) {
```
andreban: nit: missing spaces around `=`: `fetch = oneBehindFetch`

packages/core/lib/OneBehindFetch.js
sebastianbenz: done

packages/cache-list/lib/AmpCaches.js
sebastianbenz: according to eslint says it's OK :-)

PR #288, opened 2019-04-22
Always inline v0.css

sebastianbenz:
The boilerplate transformer will now inline v0.css by default. This
is safe to do as the AMP runtime will check if v0.css matches it's
current version and will update it otherwise.


sebastianbenz:
> Not sure if this 2.4k would be worth the effort though.

I agree. Thanks for the review!

PR #291, opened 2019-04-23
Valid optimized AMP

sebastianbenz:
This makes the HTML produced by AMP Optimizer valid AMP (including
server-side-rendering).

Notable changes:

* new attribute `transformed="google;v=1"` to be compatible to the AMP
cache
* log is now passed in via config => this fixes an issue where verbose
mode did not enabled correctly
* requires opt-in for now via validAmp: true


sebastianbenz:
I need to add a few more end-to-end tests ensuring that the produced output is valid AMP.

PR #293, opened 2019-04-24
Fix valid AMP transformations

sebastianbenz:
* fix runtimeVersion not being set correctly
* add valid AMP tests
* ensure validAmp flag loads only valid amp transformers


sebastianbenz:
Thanks! just verified: nextjs' tests pass as well.

PR #306, opened 2019-05-07
Optimizer redesign

sebastianbenz:
Take a look at packages/optimizer/demo/simple/index.js for the new API.


sebastianbenz:
@andreban would be great if you could take a look at the new API. There are still a few things to do:

* update the cheerio sample (maybe remove it as it's not really recommended for performance reasons).
* update optimizer-express
* ...

PR #321, opened 2019-05-20
CLI: add optimize command

sebastianbenz:


packages/cli/lib/cli.js
```diff
@@ -29,16 +30,18 @@ class Cli {
...
 
     switch (command) {
+      case 'curls':
+        return require('./cmds/curls')(args, this.logger_);
       case 'help':
         return require('./cmds/help')(args, this.logger_);
-      case 'version':
-        return require('./cmds/version')(args, this.logger_);
+      case 'optimize':
+        return new OptimizeCmd().run(args, this.logger_);
```
andreban: I'm wondering if we should make all the commands to use the same format as this. I do think it reads better than requiring inline. The downside seems to be requiring upfront makes the code to run the command load sooner, which may lead to slower "boot time" for the CLI?

sebastianbenz: Good point. If inlined requiring the command.

PR #364, opened 2019-06-21
Fix linting SXG requests

dritter:
The `dump-signedexchange` responds with `The exchange has a valid signature.`, instead of `The exchange has valid signature.`. This is why the linter returns an error.

I don't know how to test this repo (I ran into some errors). Therefore I am somewhat unsure if the verification now works correctly. So please retest this ;)

Edit: The dump-signedexchange fixed the message here: https://github.com/WICG/webpackage/commit/c7ef1c4bce81c7d1d3a60d5c17df4dcd0afb3cd3

Edit2:
On second thought, this could be easily done to be compatible with both versions of `dump-signedexchange`, if desired. Just let me know.

### DIFF
--- a/packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
+++ b/packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
@@ -33,7 +33,7 @@ export class SxgDumpSignedExchangeVerify extends Rule {
       sxg = await execa(CMD, ARGS, { input: body }).then(spawn => {
         const { stdout } = spawn;
         let m: ReturnType<typeof String.prototype.match>;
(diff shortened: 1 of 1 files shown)


dritter:
I signed it!

ithinkihaveacat:
Thanks for finding this, LGTM! @andreban are you able to merge this since @sebastianbenz is on vacation?

I think it's kinda okay to only support the newest version, shelling out to the `dump-signedexchange` binary is not going to be the most reliable thing anyway… (Note that the latest version of `dump-signexchange` supports JSON output via `-json` too, which will make this all somewhat more reliable when the linter is updated to support it.)

Issue #378, opened 2019-07-08
Create one way to handle network dependencies (validator, etc.)


ithinkihaveacat:
AMP has some dependencies on a few resources hosted on `cdn.ampproject.org`: a document is valid iff the validator hosted at `cdn.ampproject.org` says it is, the definitive source of runtime version comes from this host, etc.

For various reasons, it's not always a good idea to actually fetch these resources every time a module is invoked. (e.g. if attempting to validate multiple files it's not feasible/reasonable to fetch the validator for every file that needs to be validated; running tests should be possible without network access.)

Various AMP Toolbox modules have created different approaches to dealing with these problems, see:

- the linter, which [looks for a local file, then hits network](https://github.com/ampproject/amp-toolbox/blob/4b54f86258a28fb68a2486173331d9354fdf4734/packages/linter/src/validate.ts#L4-L7)
- runtime-version, which [uses onebehindfetch](https://github.com/ampproject/amp-toolbox/blob/4b54f86258a28fb68a2486173331d9354fdf4734/packages/runtime-version/index.js#L19-L20)
- the validator ([defaults to network](https://github.com/ampproject/amphtml/blob/b4dbb22cba0cc5255b40e47d986c7173530db4da/validator/nodejs/index.js#L304))

We should come up with a single mechanism for handling these use cases.

ithinkihaveacat:
Proposal: every module that requires runtime data takes an optional constructor argument `runtimeData` that has the signature `() => Promise<string>`.

- Having an async function in the constructor makes every method async as well. I think this will okay?
- If it were mandatory, then there would be no need for any network code in the modules themselves, which is a nice property. However, I think for ease of use it should be optional. (In which case the module will need to fetch whatever it needs to fetch if the parameter not provided.)

fstanis:
> optional constructor argument `runtimeData` that has the signature `() => Promise<string>`

I don't like this. It has several bad implications:
- The constructor can fail.
- Testing becomes hard.
- It's a leaky abstraction.

Instead, I think we should assume the data comes as a string. It's not the module's responsibility to fetch this data, it's the caller's, so we can move all the fetch logic to a util.

fstanis:
My proposal / intent to implement after discussing with @ithinkihaveacat earlier:

-   Build a utility method in `@ampproject/toolbox-core` that takes care of network lookups (see below).
[...]

PR #380, opened 2019-07-08
Use cross-fetch instead of node-fetch in core

fstanis:
This makes core isomorphic in theory (valid for both browser and node.js), which makes it easier to use from other isomorphic packages in this repo.

Related to #378.


sebastianbenz:
How does this affect our build chain?

sebastianbenz:
Closing this until we find a proper solution.

PR #389, opened 2019-07-11
Replace oneBehindFetch with cross-fetch in validator-rules

fstanis:
It seems the current solution with `oneBehindFetch` has some odd behavior here. Since this is temporary anyway (#378), just replacing it with `cross-fetch` should work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
     "@ampproject/toolbox-cache-list": {
       "version": "file:packages/cache-list",
       "requires": {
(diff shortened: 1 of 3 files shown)


sebastianbenz:
Please use `node-fetch` instead. `cross-fetch` means just more dependencies without providing any value (as long as we don't build a bundled version for the browser).

sebastianbenz:
Actually, I'm not sure if my comment is correct. Users of the library could probably create a bundle on their own.

PR #401, opened 2019-07-18
Fix "amp lint [url]"

ithinkihaveacat:
I think this fixes the broken code in packages/linter that was causing "amp lint" to generate no output, and also adds a test to prevent this happening again.

(CI likely to fail because the contents of `packages/cli/node_modules/@ampproject/*` seem to be installed directly from the npm repo (i.e. they aren't copies or symlinks to the local packages) and so even though the `toolbox-linter` module is actually updated, the tests are run against the old (published) version.)


sebastianbenz:
Closing this in favor of #402 which also fixes the project setup.

PR #415, opened 2019-07-23
Improve toolbox-linter README

ithinkihaveacat:
Improve the dev experience by clarifying how to run the npm scripts in toolbox-linter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "generate:changelog": "lerna-changelog",
     "test": "npm-run-all build test:node test:browser lint",
     "test:node": "jasmine --config=jasmine.json && cd packages/linter && npm test",
+    "test:node:linter": "cd packages/linter && npm test",
     "test:browser": "karma start --single-run --browsers ChromeHeadless karma.conf.js",
     "lint": "eslint \"**/*.js\" --ignore-path .gitignore",
     "lint:fix": "eslint \"**/*.js\" --fix --ignore-path .gitignore"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -9,6 +9,7 @@
     "generate:changelog": "lerna-changelog",
     "test": "npm-run-all build test:node test:browser lint",
     "test:node": "jasmine --config=jasmine.json && cd packages/linter && npm test",
+    "test:node:linter": "cd packages/linter && npm test",
```
sebastianbenz: can you add it to the generic `test` command as well.

packages/linter/README.md
```diff
@@ -56,11 +54,16 @@ performed.
 
 ## Development
 
-### Commands
+**Important note!** Many of the scripts below rely on binaries that are
+installed in the `../../node_modules/.bin` directory, and *will fail* if invoked
+in the default configuration. To fix this, either:
```
sebastianbenz: this feels too complicated

package.json
ithinkihaveacat: I added a `test:linter` script, is that what you meant?

packages/linter/README.md
ithinkihaveacat: Yeah, it's not great. I could just remove these scripts, but it's useful to be able to run the linter tests individually. Is there a good way around this? I don't think any of the other packages support this? (i.e. running `npm test` from the cli when in the `packages/cache-url` directory.)

sebastianbenz: Hm, I usually use the `fdescribe` feature to run specific specs while developing.

PR #420, opened 2019-07-24
amp lint now supports same features as amplint

sebastianbenz:


packages/linter/src/cli.ts
```diff
@@ -31,10 +31,10 @@ const UA = {
   ].join(" ")
 };
 
-export function cli(argv: string[], logger = console) {
+export function cli(argv: string[], logger = console, cmd = "amplint") {
   program
     // .version(version)
-    .usage(`amplint [options] URL|copy_as_cURL`)
+    .usage(`${cmd} [options] URL|copy_as_cURL`)
```
ithinkihaveacat: I don't think this is worth fixing for this PR, but I just noticed this seems to be slightly broken in both this and the previous version. commander seems to trying to guess the actual binary or something, and it often gets it wrong:

```sh
$ node bin/amp lint
Usage:  amp lint [options] URL|copy_as_cURL
```

```sh
$ node bin/amp lint -h
Usage: lint amp lint [options] URL|copy_as_cURL
```

(Invoking `amplint` directly produces similarly unexpected results.)

sebastianbenz: weird...

PR #430, opened 2019-07-31
support AMP Email Cors

sebastianbenz:


packages/cors/lib/cors.js
```diff
@@ -78,9 +84,15 @@ module.exports = (options, caches=new Caches()) => {
     }
     // Add CORS and AMP CORS headers
     response.setHeader('Access-Control-Allow-Origin', originHeaders.origin || sourceOrigin);
+    const headersToExpose = [];
     if (options.enableAmpRedirectTo) {
-      response.setHeader('Access-Control-Expose-Headers', 'AMP-Redirect-To');
+      headersToExpose.push('AMP-Redirect-To');
+    }
+    if (options.email) {
```
fstanis: Maybe add additional check to ensure it's a valid request, e.g. `if (options.email && sourceOrigin.contains('@')) {`

packages/cors/README.md
```diff
@@ -92,6 +92,19 @@ app.use(ampCors({
 }));
 ```
 
+
+### Email Mode
+
+AMP for Email has [specific AMP CORS requirements](https://developers.google.com/gmail/ampemail/security-requirements). You can enable the AMP for Email CORS mode via the `email option`:
```
fstanis: s/AMP for Email/Gmail here, since it's not standardized and will likely be replaced by

fstanis:
Looks good overall. FWIW, this is not standardized and is Gmail-specific right now: ampproject/wg-amp4email#7 is the proposal to standardize, which moves the logic to new headers instead of relying on `__amp_source_origin`.

packages/cors/lib/cors.js
sebastianbenz: Not sure if there's match value in this check...and I don't want to go down the email regex rabbit hole.

packages/cors/README.md
sebastianbenz: rephrased

sebastianbenz:
@fstanis PTAL

PR #440, opened 2019-08-05
Improve dump-signedexchange tests

ithinkihaveacat:
Check that cert-url and validity-url have a 200 status code, and have the expected content-type.

### DIFF
--- a/packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
+++ b/packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
@@ -4,6 +4,9 @@ import { fetchToCurl } from "../helper";
 import { Context } from "../index";
 import { Rule } from "../rule";
 
+const CMD_DUMP_SXG = `dump-signedexchange`;
+const CMD_DUMP_SXG_ARGS = [`-verify`, `-json`];
+
 function compare(
   expected: { [k: string]: boolean | string | number },
   actual: typeof expected
@@ -20,7 +23,7 @@ function compare(
         return acc;
       }
     },
-    {} as typeof expected
+    {} as { [k: string]: string }
   );
 }
 
@@ -41,6 +44,7 @@ const REQUEST_HEADERS = {
   "amp-cache-transform": `google;v="1"`
 };
 const EXPECTED_VERSION = "1b3";
+const SECONDS_IN_A_DAY = 24 * 60 * 60;
 
 export class SxgDumpSignedExchangeVerify extends Rule {
   async run({ url, headers }: Context) {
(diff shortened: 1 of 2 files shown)


packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
```diff
@@ -34,31 +72,65 @@ export class SxgDumpSignedExchangeVerify extends Rule {
...
-    } else {
-      return this.pass();
     }
+    return this.pass();
   }
   meta() {
     return {
       url: "",
-      title: "dump-signedexchange -verify does not report errors",
+      title: "no SXG errors found",
```
sebastianbenz: The error message: "FAIL no SXG errors found" is a bit confusing. Maybe: "valid SXG"?

packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
```diff
@@ -34,31 +72,65 @@ export class SxgDumpSignedExchangeVerify extends Rule {
...
+    if (!validityUrl) {
+      return this.fail(`Can't find valid [cert-url] [${JSON.stringify(sxg)}]`);
+    }
+    if (!(await urlHasContentType(validityUrl, headers, "application/cbor"))) {
+      return this.fail(
+        `validity-url [${validityUrl}] is not found or has wrong content type`
+      );
+    }
+    const expires = sxg.signatures[0]["Params"]["expires"] as number;
+    if (7 * 24 * 60 * 60 + Date.now() / 1000 < expires) {
```
sebastianbenz: magic number, please replace with a constant

packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
```diff
@@ -4,20 +4,57 @@ import { fetchToCurl } from "../helper";
...
+}
+
+const REQUEST_HEADERS = {
+  "accept": "application/signed-exchange;v=b3",
+  "amp-cache-transform": `google;v="1"`
+};
+const EXPECTED_VERSION = "1b3";
+
 export class SxgDumpSignedExchangeVerify extends Rule {
   async run({ url, headers }: Context) {
```
sebastianbenz: suggestion: split the method up into multiple methods, where each method implements a different check. Currently, it's very hard to determine from the code what is checked.

packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
```diff
@@ -26,6 +63,7 @@ export class SxgDumpSignedExchangeVerify extends Rule {
     const body = await res.buffer();
     const CMD = `dump-signedexchange`;
```
sebastianbenz: move constant to head

packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
```diff
@@ -26,6 +63,7 @@ export class SxgDumpSignedExchangeVerify extends Rule {
...
     const ARGS = [`-verify`, `-json`];
```
sebastianbenz: move constant to head

packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
```diff
@@ -26,6 +63,7 @@ export class SxgDumpSignedExchangeVerify extends Rule {
...
+    const DEBUG = `${fetchToCurl(url, opt, false)} | ${CMD} ${ARGS.join(" ")}`;
```
sebastianbenz: nit: lowercase as it's not a constant.

packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
sebastianbenz: suggestion: rename also to CMD_DUMP_SXG

packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
ithinkihaveacat: Changed to `verification by ${CMD_DUMP_SXG} tool`.

packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
ithinkihaveacat: Now `7 * SECONDS_IN_A_DAY`. Or you want the whole thing a constant?

packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
ithinkihaveacat: I had a few goes and couldn't figure out a way to improve it (with the current reporting structure). Have added in some whitespace to create sections and will think a bit more about it. I'm not happy with this either.

packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
ithinkihaveacat: Done.

packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
ithinkihaveacat: Done.

packages/linter/src/rules/SxgDumpSignedExchangeVerify.ts
ithinkihaveacat: Done.

ithinkihaveacat:
@sebastianbenz PTAL. I think all comments resolved with the exception of breaking the method into pieces, which I agree is needed but I'd rather not do in this PR.

sebastianbenz:
@ithinkihaveacat can you take a look at the test failure

ithinkihaveacat:
The test failure seems to be network related (and unrelated to this PR)? Although restarting it hasn't helped. Will have another look in a bit.

sebastianbenz:
works for me as well

PR #450, opened 2019-08-14
explain when boilerplate cannot be removed

sebastianbenz:


packages/optimizer/README.md
```diff
@@ -89,6 +89,12 @@ ampOptimizer.transformHtml(originalHtml, {
 })
 ```
 
+## Best Practices
+
+### Regenerate pages at least once a week
+
+AMP Optimizer inlines CSS styles required by AMP. To make sure, that the inlined CSS stays in sync with the latest AMP release, we recommend to re-generate pages at least once a week. The good news is, out-of-sync CSS will not break your page, as AMP will check the version of the inlined CSS at runtime and will automatically update it to the latest version.
```
ithinkihaveacat: Re "not break your page": it might be worth explaining what will happen in a bit more detail if the CSS is out of date--this makes it sound like there's very little downside to _not_ regenerating pages once a week. Perhaps:

> ... we recommend regenerating pages once a week. (Out-of-sync CSS will not completely break your page, but in rare circumstances it can cause AMP components to briefly appear with the "wrong" styles, such as being visible when they should be hidden. The good news is that these glitches will only be temporary, because as soon as the AMP JS starts, it will check the inlined CSS and update it if required.)

packages/optimizer/README.md
```diff
@@ -62,22 +62,22 @@ There's also a [command line version](../cli/README.md) available:
...
-AMP Optimizer inlines CSS styles required by AMP. To make sure, that the inlined CSS stays in sync with the latest AMP release, we recommend to re-generate pages at least once a week. The good news is, out-of-sync CSS will not break your page, as AMP will check the version of the inlined CSS at runtime and will automatically update it to the latest version.
+* if the`amp-experiment`, `amp-story` or `amp-dynamic-css-classes` components are used. 
+* if an AMP component uses the `media`, `sizes` or `heights` attribut ([documentation](https://amp.dev/documentation/guides-and-tutorials/learn/common_attributes/?format=websites#heights)). A simple workaround is to replace the `media`, `sizes` or `heights` attributes with normal CSS media queries.
 
-### Debugging
+* if an AMP component uses the `intrinsic` layout. The good news is: support for `intrinsic` layout is currently [work in progress](https://github.com/ampproject/amp-toolbox/issues/264). 
 
-Enable `verbose` mode to find out why the AMP boilerplate is not being removed. You can
-enable verbose mode either when creating a new optimizer instance:
+To find out, why the AMP boilerplate could not be removed, enable the `verbose`:
```
ithinkihaveacat: Perhaps reads more smoothly as: "... enable `verbose` mode`".

packages/optimizer/README.md
```diff
@@ -62,22 +62,22 @@ There's also a [command line version](../cli/README.md) available:
 $ npx @ampproject/toolbox-cli myFile.html
 ```
 
-## Best Practices
+## Why doesn't my AMP page render faster?
 
-### Regenerate pages at least once a week
+The biggest performance gain results from [removing the AMP boilerplate code](https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/server-side-rendering/#why-is-it-faster?). However, under some circumstances it's not possioble to remove the boilerplate code:
```
ithinkihaveacat: s/possioble/possible/

packages/optimizer/README.md
```diff
@@ -62,22 +62,22 @@ There's also a [command line version](../cli/README.md) available:
...
 
-AMP Optimizer inlines CSS styles required by AMP. To make sure, that the inlined CSS stays in sync with the latest AMP release, we recommend to re-generate pages at least once a week. The good news is, out-of-sync CSS will not break your page, as AMP will check the version of the inlined CSS at runtime and will automatically update it to the latest version.
+* if the`amp-experiment`, `amp-story` or `amp-dynamic-css-classes` components are used.
```
ithinkihaveacat: Maybe useful to link to the actual source here? https://github.com/ampproject/amphtml/blob/62a9eab084ccd800d80a371e2cb29cd4f9e8576a/src/render-delaying-services.js#L39-L43

packages/optimizer/README.md
sebastianbenz: makes sense

packages/optimizer/README.md
sebastianbenz: fixed

packages/optimizer/README.md

packages/optimizer/README.md

sebastianbenz:
Thanks Michael!

PR #464, opened 2019-08-28
Add script-csp toolbox package

fstanis:


packages/script-csp/lib/calculateHash.js
```diff
@@ -0,0 +1,51 @@
...
+  const data = hash.update(src);
+  const base64 = base64URLFormat(data.digest('base64'));
+  return `${algo}-${base64}`;
+}
+
+function base64URLFormat(base64) {
+  return base64
+      .replace(/=/g, '')
+      .replace(/\+/g, '-')
+      .replace(/\//g, '_');
```
sebastianbenz: can you turn this into a single regex to avoid going over the string multiple times.

packages/script-csp/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "@ampproject/toolbox-script-csp",
+  "version": "1.0.0",
```
sebastianbenz: version needs to be `1.0.1`

packages/script-csp/lib/calculateHash.js
fstanis: As discussed, the string is very short (64 characters) and this probably improves readability slightly.

packages/script-csp/README.md
```diff
@@ -0,0 +1,70 @@
...
+such as a [`Buffer`](https://nodejs.org/api/buffer.html).
+
+`options` is used to override the default options.
+
+#### Options
+
+*   `algorithm` overrides the hashing algorithm to use. Currently, the only
+    algorithm supported by `amp-script` is `sha384` which is the default value.
+
+### Example
```
sebastianbenz: Can you mention how whitespace is handledl

packages/script-csp/lib/calculateHash.js
```diff
@@ -0,0 +1,51 @@
...
+
+const crypto = require('crypto');
+
+const SUPPORTED_ALGORITHMS = new Set([
+  'sha384',
+]);
+
+const DEFAULT_ALGORITHM = 'sha384';
+
+function calculateHash(src, {algorithm = DEFAULT_ALGORITHM} = {}) {
```
sebastianbenz: please add JSDoc

fstanis: Done

packages/script-csp/lib/calculateHash.js
```diff
@@ -25,6 +25,15 @@ const SUPPORTED_ALGORITHMS = new Set([
 
 const DEFAULT_ALGORITHM = 'sha384';
 
+/**
+ * Calculates the hash of the given input and returns a valid CSP string for use
+ * with amp-script.
+ *
+ * @param {string|DataView} src - Input script
+ * @param {Object} [options] - The options object
+ * @param {string} [options.algorithm] - Algorithm to use
```
sebastianbenz: mention the default value.

packages/script-csp/README.md
fstanis: I think it's beyond scope? It just hashes the given input as-is, including whitespace. `amp-script` also expects the hash to include them and CSP works the same.

sebastianbenz: Yes, but we should mention it:

```
Here is an example on how to use the `calculateHash` function to generate an `amp-script` CSP header. It's important to note that whitespace is critical when passing a script string to `calculateHash`.
```

fstanis: Done

PR #465, opened 2019-08-28
Add AmpScriptCsp optimizer

fstanis:
Merge after #464


packages/optimizer/lib/transformers/AmpScriptCsp.js
```diff
@@ -0,0 +1,90 @@
...
+const {calculateHash} = require('@ampproject/toolbox-script-csp');
+
+/**
+ * AmpScriptCsp - adds CSP for amp-script.
+ *
+ * Currently only supports inline scripts.
+ *
+ * This transformer supports the following parameter:
+ *
+ * `append`: specifies the calculated CSPs should be appended to existing ones.
```
sebastianbenz: Do we even need this? `append=true` needs to be the default behavior] to not break existing hashes for cross-origin scripts.

packages/optimizer/lib/transformers/AmpScriptCsp.js
```diff
@@ -0,0 +1,90 @@
...
+    if (!head || !body) return;
+
+    const inlineScripts = this._findAllInlineScripts(body);
+    const hashes = new Set();
+    for (const script of inlineScripts) {
+      const content = script.children[0].data;
+      hashes.add(calculateHash(content));
+    }
+
+    const cspMeta = this._findOrCreateCspMeta(tree, head);
```
sebastianbenz: nit: I'd first add the existing hashes. This makes it clearer that optimizer added new hashes. Also, you can use `new Set(existingCsp)` instead of the loop.

packages/optimizer/lib/transformers/AmpScriptCsp.js
fstanis: I think appending by default is a bad idea from a security perspective. The idea behind CSP is to ensure that injected scripts don't work.

If an attacker is somehow able to manipulate the pre-optimizer code and add `hash('evil')` to the meta tag, the optimizer should delete this, since `<script>evil</script>` is not part of the document.

Appending is *only* useful for when someone has external cross-origin scripts, in which case they need to add these CSP hashes manually - but that's likely to be an advanced use-case. Ideally, in the future, the optimizer should handle those as well.

packages/optimizer/lib/transformers/AmpScriptCsp.js
fstanis: `new Set(existingCsp)` wouldn't work, I'd just have two sets then (and there seems to be no way to append an array to a Set). But after I moved existing hashes to the top, I was able to get rid of the loop!

packages/optimizer/lib/transformers/AmpScriptCsp.js
sebastianbenz: If `append=false` is default, then we need to disable the whole transformer by default. Otherwise, we might break existing pages with cross origin amp-script. If we disable by default, then the security advantage is not given either.

Ensuring security best practices so far hasn't been in the scope of AMP Optimizer.

fstanis: How about making it no-op by default if a `<meta name="amp-script-src" ...>` exists?

So we could have three "modes":

1.  Do nothing when CSP exists (default).
2.  Append when CSP exists.
3.  Replace when CSP exists.

packages/optimizer/lib/transformers/AmpScriptCsp.js
```diff
@@ -0,0 +1,89 @@
...
+    const html = tree.root.firstChildByTag('html');
+    const head = html.firstChildByTag('head');
+    const body = html.firstChildByTag('body');
+
+    if (!head || !body) return;
+
+    let hashes = new Set();
+
+    const cspMeta = this._findOrCreateCspMeta(tree, head);
+    if (params.append) {
```
sebastianbenz: Three things: 

1. the `append` param should be config instead which is passed via the constructor (see [here](https://github.com/ampproject/amp-toolbox/blob/master/packages/optimizer/lib/transformers/AddBlurryImagePlaceholders.js#L74) for an example). I think this should be configured on a global level and not on a per-page level. 
2. can you rename the parameter to `amp-script-csp=[append|replace]`
3. please document the parameter in the optimizer readme

sebastianbenz:
One other thing: can you please add a test for the append parameter ([example on how to configure it](https://github.com/ampproject/amp-toolbox/blob/master/packages/optimizer/spec/transformers/experimental/AmpBoilerplateTransformer/uses_given_runtime_version/input.html#L3))

packages/optimizer/lib/transformers/AmpScriptCsp.js
fstanis: Went ahead and did this. `append` is now `mode`.

sebastianbenz: perfect!

sebastianbenz:
Woohoo - thanks a lot!

PR #467, opened 2019-08-28
Keyframes transformer

### LINKED ISSUE #459: Improve KeyFramesTransformer
1. Unparsable CSS will cause transformHtml to fail with an exception. Instead, we should  log the error und leave the original CSS in place.
2. Our css parser supports compressed serialization. As we've already parsed the css, we should always re-serialize and compress the css. This should be the default, but can be disabled via an option.

sebastianbenz:
Fixes #459


packages/optimizer/lib/transformers/SeparateKeyframes.js
```diff
@@ -36,6 +43,11 @@ const allowedKeyframeProps = new Set([
 class SeparateKeyframes {
   constructor(config) {
     this.log_ = config.log.tag('SeparateKeyframes');
+    if (config.compress === false || config.compress === 'false') {
```
fstanis: Why can this be a string? Assuming that's by design, you could probably simplify with `if (String(config.compress) === 'false')`.

Could also shorten further to
```js
const compress = String(config.compress) !== 'false';
this.stringifyOptions_ = compress ? OPTIONS_COMPRESS : OPTIONS_PRETTY_PRINT;
```
and get rid of the `if` all together.

packages/optimizer/lib/transformers/SeparateKeyframes.js
```diff
@@ -125,7 +145,11 @@ class SeparateKeyframes {
     });
 
     // if no rules moved nothing to do
-    if (!keyframesTree.stylesheet.rules.length) return;
+    if (!keyframesTree.stylesheet.rules.length) {
```
fstanis: nit: `array.length === 0` is more readable than `!array.length`

packages/optimizer/lib/transformers/SeparateKeyframes.js
sebastianbenz: This was to support strings passed via the CLI, but that's the wrong place to handle it. Changed to simple boolean.

packages/optimizer/lib/transformers/SeparateKeyframes.js
sebastianbenz: done

Issue #487, opened 2019-09-09
Origin check fails when both origin and amp-same-origin headers are present


wille:
Chrome `Version 75.0.3770.100 (Official Build) (64-bit)`
Sends `origin` headers on same origin requests along with `amp-same-origin` whereas Chromium `Version 76.0.3809.132 (Official Build) Arch Linux (64-bit)` does not

From the docs
> As this request is from the same origin, there is no Origin header but the custom AMP request header of AMP-Same-Origin: true

`origin` gets picked up before the `amp-same-origin` header [here](https://github.com/ampproject/amp-toolbox/blob/master/packages/cors/lib/cors.js#L122) which fails the origin check [here](https://github.com/ampproject/amp-toolbox/blob/master/packages/cors/lib/cors.js#L78)

I'm opening a PR to check for `amp-same-origin` before `origin` which will solve this issue.

I would say it isn't expected to receive `origin` in same origin requests but [according to RFC](https://tools.ietf.org/html/rfc6454#section-7.3) it may

sebastianbenz:
Thanks for investigating! Makes sense!

PR #493, opened 2019-09-15
check 'amp-same-origin' header before 'origin' header

wille:
https://github.com/ampproject/amp-toolbox/issues/487

doesn't pick up `origin` on same origin requests where `amp-same-origin` is set


wille:
@googlebot I signed it!

wille:
express does set the incoming headers lowercased whereas the tests does not, that is why some tests are failing. work in progress.

PR #517, opened 2019-10-04
optimizer: Do not append rtv/{rtv}/ to ampUrlPrefix

mattwomple:
BREAKING CHANGE

When both `ampUrlPrefix` and `ampRuntimeVersion` are specified, do not
append `rtv/{rtv}/` to the prefix. If version information is expected in
the URL, it can be included in `ampUrlPrefix`.


sebastianbenz:
Hm. This is confusing indeed. I'd prefer to not introduce a third parameter though. My suggestion would be: don't add the runtime version if a prefix is given. If someone wants to self-host a versioned runtime, they can append the version string to the prefix. Wdyt?

mattwomple:
> Hm. This is confusing indeed. I'd prefer to not introduce a third parameter though. My suggestion would be: don't add the runtime version if a prefix is given. If someone wants to self-host a versioned runtime, they can append the version string to the prefix. Wdyt?

This would absolutely be my preference as well. It is a breaking change, though. Anyone who has optimizer implemented using `ampUrlPrefix` and `ampRuntimeVersion` with the expectation that the default pattern `/rtv/{rtv}` will automatically apply to their `ampUrlPrefix` will suddenly find their URLs broken when amp-toolbox updates in the future. Is it early enough in amp-toolbox's life that this is ok?

mattwomple:
Sorry for the force pushes. I accidentally committed changes using a different GitHub account.

mattwomple:
@sebastianbenz This breaks several tests (`npm test`) but I'm a little embarrassed to say that I don't see where the tests live. Grepping the amp-toolbox dir for failed test names didn't reveal anything. Could you point me in the right direction?

mattwomple:
Oh, nevermind:

https://github.com/ampproject/amp-toolbox/blob/6ce20613dba3e66b7722df599f62f35dc6391962/package.json#L11

Sorry for the noise.

mattwomple:
Tests have been updated. This is ready for review.

packages/optimizer/README.md
```diff
@@ -187,9 +187,13 @@ const optimizer = AmpOptimizer.create({
 
 ### Self-hosted AMP components
 
-It's possible to rewrite the AMP framework and component imports to a different domain than `cdn.ampproject.org`.
+It's possible to rewrite the AMP framework and component imports to a different domain than `cdn.ampproject.org`. These options tailor URL rewriting performed by `ampOptimizer.transformHtml`:
 
-Example:
+- `ampUrlPrefix`: Replace `https://cdn.ampproject.org/` with another origin or relative path.
+
+  **Notice:** The behavior of `ampUrlPrefix` when used in conjunction with `ampRuntimeVersion` changed beginning with version 2.0. Prior to 2.0, `rtv/{rtv}/` was automatically appended to `ampUrlPrefix` when `ampRuntimeVersion` was specified. Since version 2.0, `ampUrlPrefix` is not modified when `ampRuntimeVersion` is also specified.
```
sebastianbenz: Change 2.0 to 1.1.2 as we're not going to make a major release for this

mattwomple: Fixed in 79edabc. Thanks!

choumx:
Why not ask callers to omit `ampRuntimeVersion` if they don't want RTV paths when using `ampUrlPrefix`?

mattwomple:
> Why not ask callers to omit `ampRuntimeVersion` if they don't want RTV paths when using `ampUrlPrefix`?

Setting `ampRuntimeVersion` ensures you can maintain a correspondence between self-hosted runtime version and the version used during optimization. Consider this code: https://github.com/ampproject/amp-toolbox/blob/efe5775c5ff93ebacd6f7bd336ba02d829eca579/packages/optimizer/lib/transformers/AmpBoilerplateTransformer.js#L85-L90 . If the self-hosted runtime is pushed on Monday, and then on Wednesday an AMP page is optimized, the inline CSS in the optimized AMP page will be associated with a different runtime version.

mattwomple:
> Instead of making (3) behave like (2), why not ask callers to just do (2)?

How would optimizer know which runtime version should be used for performing its optimizations? It is not safe to assume "latest RTV available on cdn.ampproject.org". Specifying both `ampRuntimeVersion` and `ampUrlPrefix` informs optimizer what the URLs should look like to extensions and what version of boilerplate CSS should be included.

choumx:
Oh, I thought that cache transforms are not pegged to RTV. E.g. if they were, it would make AMP rollbacks very tricky. 

/cc @ampproject/wg-caching

sebastianbenz:
RTV is only relevant for v0.css. The AMP runtime will download the latest v0.css in case of a version mismatch.

choumx:
...it does? 😛 

I know runtime does this for JS (v0 vs. extension RTV mismatch) but I don't think we do this for CSS which is bundled directly in the JS.

sebastianbenz:
😀 ~~download~~ update. In any case, outdated v0.css shouldn't be an issue (at least that's what I was told that when I implemented the transformer).

PR #518, opened 2019-10-04
optimizer: Make dynamic component URL rewrites optional

### LINKED ISSUE #515: FR: Make URL rewriting of SSR AMP component optional
Feature Request

The `ampUrlPrefix` option for `ampOptimizer.transformHtml` supports rewriting all `cdn.ampproject.org` references to your own host. Some AMP components are actually server-side rendered, though. For example, if you fetch `https://cdn.ampproject.org/v0/amp-geo-0.1.js` in Spain, you get a different file than in the USA (easy to test with a VPN). Some "self-hosters" may not have the capability to do geolocation and determine what needs to update in server-side rendered JS.

It would be great if a new option could be introduced for `transformHtml`:
`rewriteSsrAmpUrls` that defaults to `true`, but could be set to `false` to preserve script URLs like `https://cdn.ampproject.org/v0/amp-geo-0.1.js` in the AMP document.

As far as I know `amp-geo` is the only component where SSR is used today. Having this option available would be beneficial to self-hosters who may not be able to keep up with the introduction of new SSR AMP components.

mattwomple:
DEPENDS ON #517

Not everyone who self-hosts the AMP runtime has the infrastructure to
support dynamic components like `amp-geo`, ref: [AMP cache guidelines](https://github.com/ampproject/amphtml/blob/master/spec/amp-cache-guidelines.md#guidelines-adding-a-new-cache-to-the-amp-ecosystem).

Introduce option rewriteDynamicComponents that prevents modification of
dynamic component URLs that would otherwise have ampUrlPrefix
substituted.

Two relevant tests added:
- rewrites_host_but_not_dynamic_components_with_rtv
- rewrites_host_but_not_dynamic_components_without_rtv

Fixes #515

### DIFF
--- a/packages/optimizer/README.md
+++ b/packages/optimizer/README.md
@@ -193,6 +193,8 @@ It's possible to rewrite the AMP framework and component imports to a different 
 
   **Notice:** The behavior of `ampUrlPrefix` when used in conjunction with `ampRuntimeVersion` changed beginning with version 1.1.2. Prior to 1.1.2, `rtv/{rtv}/` was automatically appended to `ampUrlPrefix` when `ampRuntimeVersion` was specified. Since version 1.1.2, `ampUrlPrefix` is not modified when `ampRuntimeVersion` is also specified.
 
+- `rewriteDynamicComponents`: When used in conjunction with `ampUrlPrefix`, this option can be set to `false` to prevent [dynamic AMP components](https://github.com/ampproject/amphtml/blob/master/spec/amp-cache-guidelines.md#guidelines-adding-a-new-cache-to-the-amp-ecosystem) from having their URLs rewritten.
+
 Examples:
 ```
 const ampOptimizer = require('@ampproject/toolbox-optimizer');
@@ -204,14 +206,27 @@ const originalHtml = `
 ...
 `
 
-// this will rewrite https://cdn.ampproject.org/v0.js to /amp/v0.js
+// this will rewrite https://cdn.ampproject.org/v0.js to /amp/v0.js and will
+// rewrite dynamic component https://cdn.ampproject.org/v0/amp-geo-0.1.js to
+// /amp/v0/amp-geo-0.1.js
 const optimizedHtmlA = await ampOptimizer.transformHtml(originalHtml, {
   ampUrlPrefix: '/amp'
 });
 
-// this will rewrite https://cdn.ampproject.org/v0.js to /amp/v0.js
+// this will rewrite https://cdn.ampproject.org/v0.js to /amp/v0.js and will
+// not rewrite dynamic component https://cdn.ampproject.org/v0/amp-geo-0.1.js
 const optimizedHtmlB = await ampOptimizer.transformHtml(originalHtml, {
+  ampUrlPrefix: '/amp',
+  rewriteDynamicComponents: false
+});
+
+// this will rewrite https://cdn.ampproject.org/v0.js to
+// /amp/001515617716922/v0.js and will rewrite dynamic component
+// https://cdn.ampproject.org/v0/amp-geo-0.1.js to
+// https://cdn.ampproject.org/v0/rtv/001515617716922/amp-geo-0.1.js
+const optimizedHtmlC = await ampOptimizer.transformHtml(originalHtml, {
   ampRuntimeVersion: '001515617716922',
-  ampUrlPrefix: '/amp'
+  ampUrlPrefix: '/amp/001515617716922',
+  rewriteDynamicComponents: false
 });
 ```
--- a/packages/optimizer/lib/transformers/RewriteAmpUrls.js
+++ b/packages/optimizer/lib/transformers/RewriteAmpUrls.js
@@ -15,13 +15,17 @@
  */
 'use strict';
 
-const {AMP_CACHE_HOST, appendRuntimeVersion} = require('../AmpConstants.js');
+const {
+  AMP_CACHE_HOST,
+  AMP_DYNAMIC_COMPONENTS,
+  appendRuntimeVersion,
+} = require('../AmpConstants.js');
 const {findMetaViewport} = require('../HtmlDomHelper');
 
 /**
  * RewriteAmpUrls - rewrites AMP runtime URLs.
  *
- * This transformer supports two parameters:
+ * This transformer supports several parameters:
  *
  * * `ampRuntimeVersion`: specifies a
  *   [specific version](https://github.com/ampproject/amp-toolbox/tree/master/runtime-version")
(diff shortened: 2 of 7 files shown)


mattwomple:
I meant to push this as a draft PR. Apologies.

mattwomple:
This is ready for review, with the caveat that it has been written with #517 as its base.

packages/optimizer/README.md
```diff
@@ -187,9 +187,15 @@ const optimizer = AmpOptimizer.create({
...
+It's possible to rewrite the AMP framework and component imports to a different domain than `cdn.ampproject.org`. These options tailor URL rewriting performed by `ampOptimizer.transformHtml`:
 
-Example:
+- `ampUrlPrefix`: Replace `https://cdn.ampproject.org/` with another origin or relative path.
+
+  **Notice:** The behavior of `ampUrlPrefix` when used in conjunction with `ampRuntimeVersion` changed beginning with version 2.0. Prior to 2.0, `rtv/{rtv}/` was automatically appended to `ampUrlPrefix` when `ampRuntimeVersion` was specified. Since version 2.0, `ampUrlPrefix` is not modified when `ampRuntimeVersion` is also specified.
+
+- `rewriteDynamicComponents`: When used in conjunction with `ampUrlPrefix`, this option can be set to `false` to prevent [dynamic AMP components](https://github.com/ampproject/amphtml/blob/master/spec/amp-cache-guidelines.md#guidelines-adding-a-new-cache-to-the-amp-ecosystem) from having their URLs rewritten. This option is only effective if `ampRuntimeVersion` is specified. This ensures the self-hosted runtime version matches version of the dynamic component delivered by `cdn.ampproject.org`.
```
sebastianbenz: > This option is only effective if `ampRuntimeVersion` is specified. This ensures the self-hosted runtime version matches version of the dynamic component delivered by `cdn.ampproject.org`.

I don't understand why it only makes sense in conjunction with ampRuntime Version.

packages/optimizer/lib/transformers/RewriteAmpUrls.js
```diff
@@ -35,9 +35,18 @@ const {findMetaViewport} = require('../HtmlDomHelper');
...
+ * * `rewriteDynamicComponents`: optionally disable rewriting of
+ *   [dynamically generated components](https://github.com/ampproject/amphtml/blob/master/spec/amp-cache-guidelines.md#guidelines-adding-a-new-cache-to-the-amp-ecosystem).
+ *   For example: `https://cdn.ampproject.org/v0/amp-geo-0.1.js` returns
+ *   different content depending on the country from which the request was made.
+ *
+ * All parameters are optional. If no option is provided, runtime URLs won't be
+ * re-written. You can combine `ampRuntimeVersion` and `ampUrlPrefix` to rewrite
+ * AMP runtime URLs to versioned URLs on a different origin.
+ *
+ * rewriteDynamicComponents is only effective if `ampRuntimeVersion` and
```
sebastianbenz: same as above

packages/optimizer/lib/transformers/RewriteAmpUrls.js
```diff
@@ -49,16 +58,33 @@ class RewriteAmpUrls {
...
+        ampUrlPrefix = appendRuntimeVersion(ampUrlPrefix, params.ampRuntimeVersion);
+        dynamicAmpUrlPrefix = appendRuntimeVersion(dynamicAmpUrlPrefix, params.ampRuntimeVersion);
+      } else if (params.rewriteDynamicComponents === false) {
+        dynamicAmpUrlPrefix = appendRuntimeVersion(AMP_CACHE_HOST, params.ampRuntimeVersion);
+      }
     }
 
+    // Should be kept up to date with dynamic components listed here:
+    // https://github.com/ampproject/amphtml/blob/master/spec/amp-cache-guidelines.md#guidelines-adding-a-new-cache-to-the-amp-ecosystem
+    const dynamicElements = ['amp-geo'];
```
sebastianbenz: please move this into a constant at the top of the file

packages/optimizer/lib/transformers/RewriteAmpUrls.js
```diff
@@ -49,16 +58,33 @@ class RewriteAmpUrls {
...
+    const dynamicTemplates = [];
```
sebastianbenz: same as above

packages/optimizer/lib/transformers/RewriteAmpUrls.js
```diff
@@ -49,16 +58,33 @@ class RewriteAmpUrls {
...
+
     let node = head.firstChild;
     let referenceNode = findMetaViewport(head);
 
     while (node) {
       if (node.tagName === 'script' && this._usesAmpCacheUrl(node.attribs.src)) {
-        node.attribs.src = this._replaceUrl(node.attribs.src, ampUrlPrefix);
+        const isDynamicComponent =
```
sebastianbenz: please extract this into a separate method for readability

mattwomple:
> Looks good except: I think this behavior should be supported by versioned and unversioned runtime artifacts (or am I missing anything?).

See here and my following comment: https://github.com/ampproject/amp-toolbox/issues/515#issuecomment-538893603

sebastianbenz:
Hah! Completely forgot about this 🤦‍♂️. It starts getting tricky. With the revised self-hosting behavior (`ampRuntimeVersion` no longer automatically being appended) the version will no longer be automatically enforced. However, the good news is: the AMP runtime will automatically download the matching component version if there's a mismatch. So worst case is: the extension needs to be downloaded twice, but will continue to work. That's acceptable for me. 

Hence, I'd favor making this independent of the runtime version.

packages/optimizer/lib/transformers/RewriteAmpUrls.js
mattwomple: Let me know if commit 3354b6a does this adequately.

packages/optimizer/lib/transformers/RewriteAmpUrls.js
mattwomple: Let me know if commit 3354b6a does this adequately.

packages/optimizer/lib/transformers/RewriteAmpUrls.js
mattwomple: I added it to `AMP_CONSTANTS`. Good? Bad?

packages/optimizer/lib/transformers/RewriteAmpUrls.js
mattwomple: I added it to AMP_CONSTANTS. Good? Bad?

packages/optimizer/README.md
mattwomple: This is resolved in commit 3354b6a.

packages/optimizer/lib/transformers/RewriteAmpUrls.js
mattwomple: This is resolved in commit 3354b6a.

packages/optimizer/lib/transformers/RewriteAmpUrls.js
sebastianbenz: even better

PR #522, opened 2019-10-08
Remove dependency on pika

ithinkihaveacat:


packages/linter/package.json
```diff
@@ -15,52 +15,31 @@
     "debug": "4.1.1",
     "execa": "2.0.5",
     "node-fetch": "2.6.0",
-    "probe-image-size": "5.0.0"
+    "prettier": "1.18.2",
+    "probe-image-size": "5.0.0",
+    "tap-parser": "10.0.0",
+    "typescript": "3.6.3"
```
ithinkihaveacat: These can actually be removed if the scripts are invoked via e.g. `lerna run --scope '*/toolbox-linter' lint` from the root `amp-toolbox` directory instead of `npm run lint`. (Since lerna injects the appropriate binaries into the path.)

But developers probably expect to be able to run `npm x` here (after doing an `npm install`) so I think it's okay to duplicate these binaries.

packages/linter/tests/lib.ts
```diff
@@ -24,26 +24,31 @@ export const withFixture = throat(
...
+            // wait for probe-image-size's aborts to settle
+            nockDone();
+            log(`... created fixture [${fixturePath}]`);
+            resolve(res);
+          }, 2000);
+        });
+      }
+    } catch (e) {
+      console.error(e);
+      return ({} as unknown) as T;
```
ithinkihaveacat: Wraps this block with a try ... catch. Not absolutely necessary but lessened confusing when doing some testing…

packages/linter/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
sebastianbenz: Not sure this is needed as there's one in the repo root.

packages/linter/README.md
```diff
@@ -53,54 +53,29 @@ performed.
...
+### Commands/Scripts
 
-1. Add `../../node_modules/.bin` to your path. A tool like
-   [direnv](https://direnv.net/) may make this easier.
-2. Invoke via "wrapper" scripts in `../../package.json`. "test:node:linter" is
-   an example of such a script.
+#### `prepack`
 
-### Commands
+Intended to be used when building a package. (It's automatically invoked by `npm pack`.) Populates the `dist` directory with the appropriate `*.js` files. (Tests
```
sebastianbenz: Is `npm pack` still a thing?

ithinkihaveacat: I use it to ensure that the package itself works as expected (make sure there are no bugs in `package.json` making typings unavailable to dependent modules, say). There doesn't seem to be a way to get lerna to build the packages, but not publish them--?

packages/linter/LICENSE
ithinkihaveacat: Yes, you're right looks like a LICENSE is somehow added to the packages: https://unpkg.com/browse/@ampproject/toolbox-optimizer@1.1.1/LICENSE. Maybe lerna does this? `npm pack` (which I thought lerna would be using) doesn't.

PR #569, opened 2020-01-13
Add AutoExtensionImporter

sebastianbenz:


packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,269 @@
...
+ * attributes as specified in the validation rules.
+ */
+class AutoExtensionImporter {
+  constructor(config) {
+    this.enabled = config.autoExtensionImport !== 'false' && config.autoExtensionImport !== false;
+    this.log_ = config.log.tag('AutoExtensionImporter');
+
+    // We use the validation rules to infer extension imports. The rules are downloaded once and for
+    // efficency, we initially extract all needed rules
+    this.extensionSpec = validatorRules.fetch().then((rules) => {
```
fstanis: Split this into a method (`initSpec` or so). I prefer to avoid non-trivial code in the constructor directly.

Also, consider having the constructor accept (fetched) `validatorRules` instead of fetching it yourself. This allows you to avoid using `await` every single time you access `this.extensionSpec` and makes testing easier.

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,269 @@
...
+          );
+        }
+        // Hack: fix missing attribute dependencies (e.g. amp-img => lightbox => amp-lightbox-gallery)
+        tag.attrs.filter((a) => manualAttributeToExtensionMapping.has(a.name))
+            .forEach((a) => {
+              a.requiresExtension = [manualAttributeToExtensionMapping.get(a.name)];
+            });
+
+        // Map attributes to tags and extensions (e.g. amp-img => amp-fx => amp-fx-collection)
+        tag.attrs.filter((a) => a.requiresExtension && a.requiresExtension.length > 0)
```
fstanis: `filter` + `forEach` is really hard to read IMO. Consider replacing this with a `for..of` loop instead.

The previous one is OK since your `forEach` is one line (but I'd encourage `for..of` there too).

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,269 @@
...
+            .forEach((a) => {
+              let attributeMapping = tagToAttributeMapping.get(tagName);
+              if (!attributeMapping) {
```
fstanis: This (ensuring an empty array exists as value) can be determined outside the loop rather than doing it in every iteration.

`tagToAttributeMapping.set(tagName, tagToAttributeMapping.get(tagName) || [])`

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,269 @@
...
+    });
+  }
+
+  async transform(tree, params) {
+    if (!this.enabled) {
+      return;
+    }
+    const html = tree.root.firstChildByTag('html');
+    const head = html.firstChildByTag('head');
+    if (!head) return;
```
fstanis: Can there really be no head? Does this also apply to body? Also, why not define body here too?

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,269 @@
...
+    }
+  }
+
+  findExtensionsToImportInHead_(head, extensionsToImport, existingImports) {
+    let node = head.firstChild;
+    while (node) {
+      // Detect any existing extension imports
+      const customElement = this.getCustomElement_(node);
+      if (customElement) {
+        existingImports.add(customElement);
```
fstanis: Split existing import detection into a separate method.

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,269 @@
...
+    const head = html.firstChildByTag('head');
+    if (!head) return;
+
+    // Extensions which need to be imported
+    const extensionsToImport = new Set();
+    // Keep track of existing extensions imports to avoid duplicates
+    const existingImports = new Set();
+
+    // Some AMP components need to be detected in the head (e.g. amp-access)
+    await this.findExtensionsToImportInHead_(head, extensionsToImport, existingImports);
```
fstanis: This method is not async (no harm in `await`ing it, but still).

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,269 @@
...
+      return;
+    }
+
+    // We use this for adding new import elements to the header
+    const referenceNode = findMetaViewport(head);
+
+    // Support custom runtime URLs
+    const host = calculateHost(params);
+    for (const extensionName of extensionsToImport) {
+      if (!existingImports.has(extensionName)) {
```
fstanis: Nit: I'd use `if (existingImports.has(extensionName)) { continue }`

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,269 @@
...
+        const extension = (await this.extensionSpec).extensionsMap.get(extensionName.trim());
+        this.log_.debug('auto importing', extensionName);
+        // Use the latest version by default
+        const version = extension.version[extension.version.length - 1];
```
fstanis: This makes me feel a bit uneasy, are we 100% sure the latest is the last in the array?

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,269 @@
...
+      if (node.tagName) {
+        promises.push(this.addRequiredExtensionByTag_(extensionsToImport, node));
+        promises.push(this.addRequiredExtensionByAttributes_(extensionsToImport, node));
+      }
+      node = node.nextNode();
+    }
+    return Promise.all(promises);
+  }
+
+  async addRequiredExtensionByTag_(allRequiredExtensions, node) {
```
fstanis: The argument order is reversed here (output set is first, while for `findExtensionsToImportInBody_` output set is last). I prefer first, but anything works as long as it's consistent.

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,269 @@
...
+    let node = body;
+    const promises = [];
+    while (node !== null) {
+      if (node.tagName) {
+        promises.push(this.addRequiredExtensionByTag_(extensionsToImport, node));
+        promises.push(this.addRequiredExtensionByAttributes_(extensionsToImport, node));
+      }
+      node = node.nextNode();
+    }
+    return Promise.all(promises);
```
fstanis: Nit: unless you need the result of these promises (I think you don't?), replace with `await Promise.all(promises)` and return nothing.

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,269 @@
...
+
+  async addRequiredExtensionByAttributes_(allRequiredExtensions, node) {
+    if (!node.tagName || !node.attribs) {
+      return;
+    }
+    // Look for element attributes indicating AMP components (e.g. amp-fx)
+    const extensionSpec = await this.extensionSpec;
+    const tagToAttributeMapping = extensionSpec.tagToAttributeMapping;
+    const attributesForTag = tagToAttributeMapping.get(node.tagName);
+    if (attributesForTag) {
```
fstanis: Nit: If you add `|| []` to the line above, you won't need this `if`.

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,269 @@
...
+    // true if we need to import amp-bind
+    let usesAmpBind = false;
+    for (const attributeName of attributeNames) {
+      if (!attributeName.startsWith(BIND_SHORT_FORM_PREFIX)) {
+        continue;
+      }
+      const attributeNameWithoutBindPrefix =
+        attributeName.substring(BIND_SHORT_FORM_PREFIX.length);
+
+      // Rename attribute from bindx to data-amp-bind-x
```
fstanis: Out of curiosity, why?

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,269 @@
...
+      if (ampBindAttrs.has(attributeNameWithoutBindPrefix)) {
+        const newAttributeName = `data-amp-bind-${attributeNameWithoutBindPrefix}`;
```
fstanis: Nit: I'd move `data-amp-bind-` to a constant.

packages/optimizer/lib/RuntimeHostHelper.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+'use strict';
+
+const {
+  AMP_CACHE_HOST,
+  appendRuntimeVersion,
+} = require('./AmpConstants.js');
+
+function calculateHost(params) {
+  let ampUrlPrefix = params.ampUrlPrefix || AMP_CACHE_HOST;
```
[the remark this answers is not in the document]
sebastianbenz: very nice!

packages/optimizer/lib/transformers/AutoExtensionImporter.js
sebastianbenz: VS Code does not autocomplete `data-X` attributes. With `bindX` you get component specific autocompletion for bindable attributes.

packages/optimizer/lib/transformers/AutoExtensionImporter.js
sebastianbenz: done

sebastianbenz:
@fstanis great review! thanks a lot! PTAL

packages/optimizer/lib/RuntimeHostHelper.js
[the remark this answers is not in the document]
fstanis: You can just check `ampUrlPrefix !== AMP_CACHE_HOST` instead of `!ampUrlPrefix`. But your call, this is fine too.

packages/optimizer/lib/transformers/AutoExtensionImporter.js
fstanis: Hm, I think the overhead is practically non-existent (looping a list of ~20 items twice). But if you want to keep it this way, rename the method to something like `parseHead_` since it's not (just) finding extensions to import (unlike its body counterpart). The current name confused me when I was reading this code.

morsssss:
@fstanis , yes, this looks so thorough that I don't feel obliged to take a look myself!

As long as you've looked a bit at https://github.com/garanj/amphtml-autoscript to make sure you don't repeat any of our mistakes :)

packages/optimizer/lib/transformers/AutoExtensionImporter.js
fstanis: That works. FWIW, the caller still has the same ability to decide: `async` methods will always return a promise that is resolved when everything inside is `await`ed. In other words, the only difference between `await Promise.all(...)` with no return and `return Promise.all(...)` is that the latter returns a promise that resolves to an array while the former resolves to `undefined` (which is more sensible here). In other words, it'd be like `return Promise.all(...).then(_ => void(0))`

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,288 @@
...
+    this.initExtensionSpec_(config.validatorRules);
+  }
+
+  /**
+   * @private
+   */
+  async initExtensionSpec_(validatorRules) {
+    this.extensionSpec = validatorRules.fetch().then((rules) => {
+      // Map extension names to more info
+      const extensionsMap = new Map(rules.extensions.map((ext) => [ext.name, {
```
fstanis: Thinking about this... you may want to take `htmlFormat` into account. For example, AMP for Email doesn't support `amp-carousel` 0.2.

More importantly, this assumes there's exactly one extension for a given `ext.name` which is incorrect because of the above. I'm not sure what the current behavior is, but it will give you an arbitrary `amp-carousel` that may be the email one (and thus 0.1) or the regular one (and thus 0.1 and 0.2).

sebastianbenz: Yup. That's already on my todo list.

packages/optimizer/lib/transformers/AutoExtensionImporter.js
sebastianbenz: renamed the method

packages/optimizer/lib/RuntimeHostHelper.js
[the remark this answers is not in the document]
sebastianbenz: done

sebastianbenz:
Forgot to mention earlier, the test runner automatically picks up tests (every folder is a separate test with input and expected output).

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -0,0 +1,307 @@
...
+  /**
+   * @private
+   */
+  async initExtensionSpec_(validatorRules) {
+    this.extensionSpec = validatorRules.fetch().then((rules) => {
+      // Map extension names to info required for generating the extension imports
+      const extensionsMap = new Map();
+      for (const ext of rules.extensions) {
+        if (ext.htmlFormat.includes(DEFAULT_FORMAT)) {
+          extensionsMap.set(ext.name, {
```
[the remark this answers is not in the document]
sebastianbenz: This was actually a bug: `DEFAULT_FORMAT` needs to be `this.format`. Throwing an exception here feels like overkill, this would mean if the validation format has a bug it'll mean pages will suddenly start to break which we should avoid if possible.

sebastianbenz:
Thanks everyone for the feedback!

PR #580, opened 2020-01-23
Fix: ReorderHead priorities

### LINKED ISSUE #579: Optimizer: ReorderHead has different priorities than in the Go version
As already discussed, I consider the Go version of the optimizer the canonical one.

The `ReorderHead` in the JS package puts things into a different order than the Go one. As I'm trying to run both the Go manual tests as well as the JS spec files, this leads to inconsistencies and test failures.

Go:
```go
// ReorderHead reorders the children of <head>. Specifically, it
// orders the <head> like so:
// (0) <meta charset> tag
// (1) <style amp-runtime> (inserted by ampruntimecss.go)
// (2) remaining <meta> tags (those other than <meta charset>)
// (3) AMP runtime .js <script> tag
// (4) AMP viewer runtime .js <script> tag
// (5) <script> tags that are render delaying
// (6) <script> tags for remaining extensions
// (7) <link> tag for favicons
// (8) <link> tag for resource hints
// (9) <link rel=stylesheet> tags before <style amp-custom>
// (10) <style amp-custom>
// (11) any other tags allowed in <head>
// (12) AMP boilerplate (first style amp-boilerplate, then noscript)
```

NodeJS:
```js
/**
 * ReorderHeadTransformer - This DOM transformer reorders tags
 * within the head. Specifically, it orders the head like so:
 * (0) meta charset, then remaining meta tags.
 * (1) <style amp-runtime> (inserted by ServerSideRenderingTransformer)
 * (2) AMP runtime .js <script> tag
 * (3) <script> tags for render delaying extensions
 * (4) <script> tags for remaining extensions
 * (5) <link> tag for favicon
 * (6) <link rel=stylesheet> tags before <style amp-custom>
 * (7) <style amp-custom>
 * (8) any other tags allowed in <head>
 * (9) amp boilerplate (first style amp-boilerplate, then noscript).
 */
```

Is the order of the Go package canonical? If so, why does the JS package diverge? If not, is the Go version out-of-date?

sebastianbenz:
ReorderHead has different priorities than in the Go version (Fixes #579)

### DIFF
--- a/packages/optimizer/lib/transformers/ReorderHeadTransformer.js
+++ b/packages/optimizer/lib/transformers/ReorderHeadTransformer.js
@@ -60,12 +60,12 @@ class HeadNodes {
     appendChild(head, this._metaCharset);
     appendChild(head, this._styleAmpRuntime);
     appendChild(head, this._linkStyleAmpRuntime);
-    appendAll(head, this._resourceHintLinks);
     appendAll(head, this._metaOther);
     appendChild(head, this._scriptAmpEngine);
     appendAll(head, this._scriptRenderDelayingExtensions);
     appendAll(head, this._scriptNonRenderDelayingExtensions);
     appendAll(head, this._linkIcons);
+    appendAll(head, this._resourceHintLinks);
     appendAll(head, this._linkStylesheetsBeforeAmpCustom);
     appendChild(head, this._styleAmpCustom);
     appendAll(head, this._others);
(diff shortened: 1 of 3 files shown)


sebastianbenz:
@schlessera PTAL

schlessera:
@sebastianbenz Looks good already, but the `AMP viewer runtime .js <script> tag` one seems to be missing still (N° 4 for Go). This check seems to be completely missing from the Node.JS version.

sebastianbenz:
@schlessera that's intentional. The viewer JS is only required if the AMP page is served inside a viewer, which is not the case for AMPs transformed by Optimizer. Theoretically, we could leave it in - but I didn't consider it needed when I implemented the transformer.

schlessera:
As I'm running the both the tests from the Go version as well as the spec test suite in here, the above was an actual blocker. The viewer JS is not, as the spec test suite doesn't even include it in the first place. So I'll just implement that to get the Go tests to pass and then this is good from my side.

packages/optimizer/lib/transformers/ReorderHeadTransformer.js
```diff
@@ -60,12 +60,12 @@ class HeadNodes {
     appendChild(head, this._metaCharset);
     appendChild(head, this._styleAmpRuntime);
     appendChild(head, this._linkStyleAmpRuntime);
```
schlessera: `styleAmpRuntime` & `linkStyleAmpRuntime` order needs to be swapped as well.

See: https://github.com/ampproject/amppackager/blob/581d266a49d13384679697031cdd2a921f454d0c/transformer/transformers/reorderhead.go#L94-L99

sebastianbenz: the order is irrelevant for these two as they are mutually exclusive (amp-runtime is the inlined version of v0.css).

sebastianbenz: Changed the order for consistency and to avoid future confusion.

PR #581, opened 2020-01-23
Fix: Empty <amp-experiment> blocks boilerplate removal

### LINKED ISSUE #578: Empty <amp-experiment> blocks boilerplate removal
The check for `<amp-experiment>` in the `ServerSideRendering` transformer is missing a check to make sure the `<amp-experiment>` is not just empty.
https://github.com/ampproject/amp-toolbox/blob/2a81fc3795a0a5354920f7c2855af9900eecacb6/packages/optimizer/lib/transformers/ServerSideRendering.js#L88

In the Go implementation, you can see an additional `isAmpExperimentUsed()` check: https://github.com/ampproject/amppackager/blob/v2/transformer/transformers/serversiderendering.go#L127

I'm currently facing the issues that the spec files seem to be wrong as this check is missing, so my tests are failing here:
https://travis-ci.org/ampproject/amp-wp/jobs/640983826#L1270-L1340

sebastianbenz:
Fixes #578


sebastianbenz:
@schlessera PTAL

packages/optimizer/lib/transformers/ServerSideRendering.js
```diff
@@ -85,7 +85,7 @@ class ServerSideRendering {
       // amp-experiment is a render delaying extension iff the tag is used in
       // the doc. We check for that here rather than checking for the existence
       // of the amp-experiment script in IsRenderDelayingExtension below.
-      if (node.tagName === 'amp-experiment') {
+      if (node.tagName === 'amp-experiment' && !this.isAmpExperimentUsed(node)) {
```
schlessera: The logic is the wrong way around:
```suggestion
      if (node.tagName === 'amp-experiment' && this.isAmpExperimentUsed(node)) {
```

sebastianbenz:
Hah! Fooled myself there. Thanks for noticing. Should be fixed now. I've also added more tests and improved the test names. @schlessera PTAL

packages/optimizer/lib/transformers/ServerSideRendering.js
sebastianbenz: done

schlessera:
LGTM!

sebastianbenz:
Thanks for the review!

PR #588, opened 2020-01-28
Add MinifyHtml transformer

sebastianbenz:
* collapses whitespace (except in <pre>, <textarea>, <script>
* trims whitespace (except in <pre>, <textarea>)
* removes comments except if they match a given pattern
* minifies JSON
* minifies inline amp-script using terser


packages/optimizer/lib/transformers/MinifyHtml.js
```diff
@@ -0,0 +1,174 @@
...
+      child.data = result.code;
+    }
+  }
+
+  minifyJson(child) {
+    try {
+      child.data = JSON.stringify(JSON.parse(child.data), null, '');
+    } catch (e) {
+      // invalid JSON
+      this.log.warn('Invalid JSON', child.data);
```
andreban: If the JSON is invalid, does it still parse in the browser? Just wondering if this somehow fail more explicitly if that's the case (or maybe not in the scope of optimizer?)

packages/optimizer/lib/transformers/MinifyHtml.js
```diff
@@ -0,0 +1,172 @@
...
+const WHITESPACE_REGEX = /\s+/g;
+
+/**
+ * MinifyHtml - minifies files size by:
+ *
+ * - minifying inline JSON
+ * - minifying inline amp-script using https://www.npmjs.com/package/terser
+ * - collapsing whitespace outside of pre, script, style and area.
+ * - removing comments
+ *
```
schlessera: Here's a list of additional transformations you could consider:
- removing quotes where they are not needed (`lang="en_US"` => `lang=en_US`)
- removing empty attribute values (`data-bool-flag=""` => `data-bool-flag`)
- removing unneeded attribute values (`disabled="disabled"` => `disabled`)
- removing redundant attribute values where the values is the default (`<input type=number step=1>` => `<input type=number>`)
- removing optional tags: https://html.spec.whatwg.org/multipage/syntax.html#optional-tags
- decoding HTML entities into Unicode characters (`&copy;` => `©`)

Probably not needed within an Amp context:
- removing `type="text/javascript"` from `<script>` tags
- removing `type="text/css"` from `<style>` tags

packages/optimizer/lib/transformers/MinifyHtml.js
```diff
@@ -0,0 +1,172 @@
...
+      node.attribs.target === 'amp-script';
+  }
+
+  isJson(node) {
+    return node.attribs &&
+      (node.attribs.type === 'application/json' || node.attribs.type === 'application/ld+json');
+  }
+
+  canCollapseWhitespace(tagName) {
+    return !/^(?:script|style|pre|textarea)$/.test(tagName);
```
schlessera: Not sure how important performance is here, but as this runs over every loop, it might be worthwhile to test using direct string comparisons here instead of a regex.

packages/optimizer/lib/transformers/MinifyHtml.js
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  canTrimWhitespace(tagName) {
+    return !/^(?:pre|textarea)$/.test(tagName);
```
schlessera: Not sure how important performance is here, but as this runs over every loop, it might be worthwhile to test using direct string comparisons here instead of a regex.

packages/optimizer/lib/transformers/MinifyHtml.js
```diff
@@ -0,0 +1,172 @@
...
+    }
+    if (opts.commentIgnorePattern.test(node.data)) {
+      return;
+    }
+    nodesToRemove.push(node);
+  }
+
+  minifyScriptNode(node, opts) {
+    const isJson = this.isJson(node);
+    const isAmpScript = this.isInlineAmpScript(node);
```
schlessera: Shouldn't one of these exclude the other? No need to process the second if the first is already true.

packages/optimizer/lib/transformers/MinifyHtml.js
```diff
@@ -0,0 +1,172 @@
...
+const COMMENT_DEFAULT_IGNORE = /^\s*__[a-bA-Z0-9_-]+__\s*$/;
+
+// Matches all consecutive whitesapce
+const WHITESPACE_REGEX = /\s+/g;
+
+/**
+ * MinifyHtml - minifies files size by:
+ *
+ * - minifying inline JSON
+ * - minifying inline amp-script using https://www.npmjs.com/package/terser
```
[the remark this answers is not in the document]
sebastianbenz: This currently happens in the SeparateKeyframes as we're already parsing the CSS there.

packages/optimizer/lib/transformers/MinifyHtml.js
sebastianbenz: Thanks for the suggestions! I've filed https://github.com/ampproject/amp-toolbox/issues/594

Some things are already handled by the parser (e.g. removing empty attribute values)

packages/optimizer/lib/transformers/MinifyHtml.js
sebastianbenz: just tested this, non-regex is 25% faster. Addressed in https://github.com/ampproject/amp-toolbox/pull/591/commits/62c343e5685670d67a5fd3e0ff348bf0867d3c89

packages/optimizer/lib/transformers/MinifyHtml.js
sebastianbenz: addressed in https://github.com/ampproject/amp-toolbox/pull/591/commits/62c343e5685670d67a5fd3e0ff348bf0867d3c89

packages/optimizer/lib/transformers/MinifyHtml.js
sebastianbenz: addressed in https://github.com/ampproject/amp-toolbox/pull/591/commits/62c343e5685670d67a5fd3e0ff348bf0867d3c89

PR #591, opened 2020-01-28
Only trim whitespace in head

sebastianbenz:


packages/optimizer/lib/transformers/MinifyHtml.js
```diff
@@ -77,8 +75,11 @@ class MinifyHtml {
     if (opts.canCollapseWhitespace && !this.canCollapseWhitespace(node.tagName)) {
       childOpts.canCollapseWhitespace = false;
     }
-    if (opts.canTrimWhitespace && !this.canTrimWhitespace(node.tagName)) {
-      childOpts.canTrimWhitespace = false;
+    if (node.tagName === 'head' || node.tagName === 'html') {
+      childOpts.inBody = false;
+    }
+    if (node.tagName === 'body') {
```
schlessera: This should be an `else if`, this is mutually exclusive to the previous `if`.

schlessera: The entire `isBody` logic to decide whether to trim would fail if there is no `<body>` tag in the document. Is it assumed the `<body>` tag exists because of previous transformers?

packages/optimizer/lib/transformers/MinifyHtml.js
```diff
@@ -77,8 +75,11 @@ class MinifyHtml {
     if (opts.canCollapseWhitespace && !this.canCollapseWhitespace(node.tagName)) {
       childOpts.canCollapseWhitespace = false;
     }
-    if (opts.canTrimWhitespace && !this.canTrimWhitespace(node.tagName)) {
-      childOpts.canTrimWhitespace = false;
+    if (node.tagName === 'head' || node.tagName === 'html') {
```
schlessera: What about the `<!doctype>` tag?

packages/optimizer/spec/transformers/valid/MinifyHtml/minifies_inline_JSON/expected_output.html
```diff
@@ -1,4 +1,5 @@
...
+
```
schlessera: Shouldn't blank lines outside of the `<body>` be trimmed as well?

packages/optimizer/lib/transformers/MinifyHtml.js
sebastianbenz: done

sebastianbenz: Yes, the AddMandatoryTag transformer will add the body.  Otherwise, I'd prefer to err on the save side and not trim whitespaces.

packages/optimizer/lib/transformers/MinifyHtml.js
sebastianbenz: Afaik, it can't have children

packages/optimizer/spec/transformers/valid/MinifyHtml/minifies_inline_JSON/expected_output.html
sebastianbenz: Missed that. Fixed that on the parser level (trim doc before parsing).

sebastianbenz:
@andreban `<p>Hello<span> World</span></p>` results in "HelloWorld" when trimming in body. The algorithm needs to be more sophisticated and check if there's a preceding whitespace.

packages/optimizer/spec/transformers/valid/MinifyHtml/minifies_inline_JSON/expected_output.html
sebastianbenz: however, spaces after doctype and before `html` still won't be removed.

PR #614, opened 2020-02-12
Update readme and add Markdown sample

sebastianbenz:


packages/optimizer/README.md
```diff
@@ -35,38 +36,165 @@ const AmpOptimizer = require('@ampproject/toolbox-optimizer');
...
 ```
 
-Checkout [the samples](demo/simple/index.js) to learn how to customize AMP Optimizer.
-
 ## Why doesn't my AMP page render faster?
 
 The biggest performance gain results from [removing the AMP boilerplate code](https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/server-side-rendering/#why-is-it-faster?). However, under some circumstances it's not possible to remove the boilerplate code:
 
 * if the`amp-experiment`, `amp-story` or `amp-dynamic-css-classes` components are used ([code](https://github.com/ampproject/amphtml/blob/62a9eab084ccd800d80a371e2cb29cd4f9e8576a/src/render-delaying-services.js#L39-L43)).
 * if an AMP component uses the `media`, `sizes` or `heights` attribut ([documentation](https://amp.dev/documentation/guides-and-tutorials/learn/common_attributes/?format=websites#heights)). A simple workaround is to replace the `media`, `sizes` or `heights` attributes with normal CSS media queries.
```
CrystalOnScript: `attribut` --> `attribute`

PR #621, opened 2020-02-14
add github action for cross os testing

sebastianbenz:


mdmower:
I noticed Git on Windows chokes on `packages/optimizer/spec/transformers/valid/ServerSideRendering/boilerplate_not_removed_when_amp_experiment_present`, presumably because a file and folder cannot have the same name within a directory. Could we fix this problem independent of this PR so that it is possible to clone and checkout branches on Windows?

sebastianbenz:
Hm. I don't quite understand. Currently, there are no files in https://github.com/ampproject/amp-toolbox/tree/master/packages/optimizer/spec/transformers/valid/ServerSideRendering. What am I missing?

mdmower:
One directory deeper than the link you pasted: https://github.com/ampproject/amp-toolbox/tree/master/packages/optimizer/spec/transformers/valid/ServerSideRendering/boilerplate_not_removed_when_amp_experiment_present

Most filesystems balk at file and folder with the same name in a directory ([example](https://unix.stackexchange.com/questions/22447/why-cant-i-have-a-folder-and-a-file-with-the-same-name)). Looking more carefully, maybe the reason we aren't seeing issues on *nix is the folder name actually has a trailing space:
```
mdmower@X1 ~/source/amp-toolbox/packages/optimizer/spec/transformers/valid/ServerSideRendering/boilerplate_not_removed_when_amp_experiment_present $ ll
total 20
drwxr-xr-x  3 mdmower mdmower 4096 Feb 15 12:50  ./
drwxr-xr-x 26 mdmower mdmower 4096 Feb 15 12:50  ../
-rw-r--r--  1 mdmower mdmower 1413 Feb 15 12:50  expected_output.html
-rw-r--r--  1 mdmower mdmower 1364 Feb 15 12:50  input.html
drwxr-xr-x  2 mdmower mdmower 4096 Feb 15 12:50 'input.html '/
```

PR #622, opened 2020-02-16
optimizer: Add meta tags for singleDoc self-host

mdmower:
Add meta tags to AMP documents to support self-hosted AMP runtime
support when AMP pages are viewed in single-doc mode.


mdmower:
Corresponding amphtml PR: https://github.com/ampproject/amphtml/pull/26829
Self-hosted runtime sample project: https://github.com/mdmower/amp-self-host-demo/

PR #636, opened 2020-02-29
optimizer: Output amp-geo API meta tags

mdmower:
- `transformHtml` paramter `geoApiUrl` enables amp-geo to use a fallback API
  if the AMP hosting environment does not support dynamically patching
  `{{AMP_ISO_COUNTRY_HOTPATCH}}` at delivery time
- Move meta tags added by transformer before all scripts to ensure their
  content is available by the time scripts run

Depends on https://github.com/ampproject/amphtml/pull/26829


sebastianbenz:
Looks good. One suggestion: update the [ReorderHeadTransformer](https://github.com/ampproject/amp-toolbox/blob/master/packages/optimizer/lib/transformers/ReorderHeadTransformer.js) to move the meta tag to the beginning of. the head.

honeybadgerdontcare:
@sebastianbenz The CSS should always be first. Why is this tag more important than the CSS?

sebastianbenz:
Forget my comment. ReorderHeadTransformer does already the right thing and adds all `meta` tags before all `script` tags

https://github.com/ampproject/amp-toolbox/blob/master/packages/optimizer/lib/transformers/ReorderHeadTransformer.js#L63

mdmower:
@sebastianbenz This PR has been moved out of draft status and is ready for any additional review comments you might have. Thanks!

mdmower:
> @sebastianbenz The CSS should always be first. Why is this tag more important than the CSS?

The meta tags added by this PR are inserted just before the first `<script>` in the head, so I don't think changes are necessary here. It would probably be worthwhile to open a new issue to ensure Optimizer orders head tags correctly in general, though.

PR #653, opened 2020-03-19
Don't separate keyframes for AMP Stories

sebastianbenz:


sebastianbenz:
Thanks for the review!

PR #679, opened 2020-04-01
Update to use postcss and cssnano for SeparateKeyframes transform

ijjk:
This updates the SeparateKeyframes transform to leverage `postcss` instead of the `css` package. It also adds `cssnano` to provide better minification when enabled.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -321,7 +321,7 @@
         "domutils": "2.0.0",
         "htmlparser2": "4.1.0",
         "normalize-html-whitespace": "1.0.0",
-        "postcss": "^7.0.27",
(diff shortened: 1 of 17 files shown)


sebastianbenz:
Thanks for doing this. Can you please rebase against master, this should fix the build error.

ijjk:
@sebastianbenz I double checked and it looks like this branch is caught up with master correctly. The failing test appears to also be [failing for master](https://travis-ci.org/github/ampproject/amp-toolbox/builds/669738263?utm_source=github_status&utm_medium=notification) right now. Let me know if I'm missing something and I can update this branch 👍

sebastianbenz:
My mistake, fix was not yet on master (but is now).

PR #683, opened 2020-04-02
Rename download-runtime => runtime-fetch

sebastianbenz:
NPM doesn't allow modules containing `download`.


packages/runtime-fetch/lib/DownloadRuntime.js
```diff
@@ -19,7 +19,7 @@ const cacheListProvider = require('@ampproject/toolbox-cache-list');
 const crossFetch = require('cross-fetch');
 const fse = require('fs-extra');
 const https = require('https');
-const log = require('@ampproject/toolbox-core').log.tag('AMP Download Runtime');
+const log = require('@ampproject/toolbox-core').log.tag('Runtime Fetch');
```
mdmower, line 22: nit: Tag doesn't match tool name; this is the only place "Runtime Fetch" is user facing.

sebastianbenz: renamed to `Runtime Download`

sebastianbenz:
Thanks for the review!

PR #691, opened 2020-04-05
optimizer: Add lts option to URL rewriter

### LINKED ISSUE #629: Feature request: Support LTS as runtime version in AMP Optimizer
See https://github.com/ampproject/amphtml/blob/master/contributing/lts-release.md

/cc @sebastianbenz @nainar

mdmower:
* Support optimizing AMP pages with long-term stable URLs:
  https://github.com/ampproject/amphtml/blob/master/contributing/lts-release.md
* Update CLI with support for all current rewrite URLs options
* Add/Update URL rewrite and boilerplate CSS tests

Note that lts option is incompatible with specifying a specific runtime version in URL rewrites. An error will be thrown if these options are used together.

Depends on #695
Fixes #629


mdmower:
Not quite ready. Need to update CSS inliner to use `ltsCssUrl` from https://cdn.ampproject.org/rtv/metadata.

mdmower:
Depends on #695
Ready for review.

packages/optimizer/lib/transformers/AmpBoilerplateTransformer.js
```diff
@@ -84,36 +84,20 @@ class AmpBoilerplateTransformer {
...
-      } else {
-        version = await this.runtimeVersion_.currentVersion({ampUrlPrefix});
-        const customCssUrl = ampUrlPrefix + V0_CSS_PATH;
-        if (this._isAbsoluteUrl(customCssUrl)) {
-          v0CssUrl = customCssUrl;
-        }
-      }
+    // TODO: If ampUrlPrefix is a relative URL, this will fall back to
+    // fetching the latest runtime version and boilerplate CSS from
+    // cdn.ampproject.org. Is this our best option?
```
sebastianbenz, line 91: Another option would be to support passing v0.css via a parameter

mdmower: Thanks for the review. I left this fallback in place so that this PR doesn't affect handling of relative `ampUrlPrefix`. I think we're both on-board with using absolute URLs for `ampUrlPrefix` going forward, so this will disappear in a future PR. Can we keep this as-is just for the purposes of this PR?

sebastianbenz: absolutely. thanks!

PR #695, opened 2020-04-05
runtime-version: Update for custom hosts and lts

mdmower:
* Support `lts` flag, both as an argument to `.getVersion()` and as a
  CLI argument.
* Fall back to `/version.txt` if `/rtv/metadata` is not available.
  This feature is only available for standard `prod` lookups, it
  is not used when a `canary` or `lts` release is requested. This
  lessens the requirements for self-hosting the runtime.
* Remove RTV padding. There's little chance that padding
  the result is going to ever result in a valid RTV.
* Gracefully return `undefined` (or empty string in CLI) when version
  is not available.
* Update test spec to use fetch-mock so network requests are not made.

### DIFF
--- a/packages/runtime-version/lib/RuntimeVersion.js
+++ b/packages/runtime-version/lib/RuntimeVersion.js
@@ -19,33 +19,52 @@ const log = require('@ampproject/toolbox-core').log.tag('AMP Runtime Version');
 
 const AMP_CACHE_HOST = 'https://cdn.ampproject.org';
 const RUNTIME_METADATA_PATH = '/rtv/metadata';
+const VERSION_TXT_PATH = '/version.txt';
 
 /**
- * Queries https://cdn.ampproject.org/rtv/metadata for the latest AMP runtime version. Uses a
+ * @typedef {number} ReleaseType
+ */
+
+/**
+ * Release type "enumeration"
+ *
+ * @enum {ReleaseType}
+ */
+const ReleaseType = {
+  canary: 0,
+  prod: 1,
+  lts: 2,
+};
+
+/**
+ * Queries <host>/rtv/metadata for the latest AMP runtime version. If version is not available from
+ * this endpoint, falls back to <host>/version.txt and manually prepends config code. Uses a
  * stale-while-revalidate caching strategy to avoid refreshing the version.
  *
- * More details: https://cdn.ampproject.org/rtv/metadata returns the following metadata:
+ * /rtv/metadata endpoint details:
  *
  * <pre>
  * {
- *    "ampRuntimeVersion": "CURRENT_PROD",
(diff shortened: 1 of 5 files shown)


mdmower:
Here are some samples of runtime version supported added by this PR:

AMP Project LTS
```
$ packages/cli/bin/amp runtime-version --lts
> 012002251816300
```

Self-hosted runtime with custom version
```
$ packages/cli/bin/amp runtime-version --host "https://ampdemo.cmphys.com/amp-rt"
> 01v1.2a
```

Self-hosted canary runtime with custom version
```
$ packages/cli/bin/amp runtime-version --host "https://ampdemo.cmphys.com/amp-rt" --canary
> 00v1.2a
```

Bing does not have /rtv/metadata
```
$ packages/cli/bin/amp runtime-version --host "https://www.bing-amp.com"
> 011912201827130
```

Bing does not have canary releases
```
$ packages/cli/bin/amp runtime-version --host "https://www.bing-amp.com" --canary
> AMP Runtime Version ERROR RTV could not be verified
```

mdmower:
With respect to bullet point "Remove unnecessary RTV padding. There is nothing in amphtml that
prevents a version from being defined by...", I've decided to open a question at amphtml: https://github.com/ampproject/amphtml/issues/27579

mdmower:
> With respect to bullet point "Remove unnecessary RTV padding. There is nothing in amphtml that
> prevents a version from being defined by...", I've decided to open a question at amphtml: [ampproject/amphtml#27579](https://github.com/ampproject/amphtml/issues/27579)

We only have to worry about 15-digit RTVs. I still question any benefit of padding a result from `/rtv/metadata` that isn't 15-digits. It is very unlikely that padding would identify a "correct" version. It seems sensible to remove the padding, so I'm leaving the PR as is.

sebastianbenz:
> We only have to worry about 15-digit RTVs. I still question any benefit of padding a result from /rtv/metadata that isn't 15-digits. It is very unlikely that padding would identify a "correct" version. It seems sensible to remove the padding, so I'm leaving the PR as is.

The padding is a leftover from the time before we had a /rtv/metadata endpoint.

packages/runtime-version/lib/RuntimeVersion.js
```diff
@@ -56,49 +75,153 @@ class RuntimeVersion {
...
+      return;
+    }
+
+    let versionTxtUrl = host + VERSION_TXT_PATH;
+    log.debug(`Falling back to ${versionTxtUrl}`);
+
+    let response;
+    try {
+      response = await this.fetch_(versionTxtUrl);
+    } catch (ex) {}
```
sebastianbenz, line 186: please add a comment explaining why we can ignore the exception

mdmower: Comment added. Thanks for reviewing!

sebastianbenz:
Thanks!

PR #698, opened 2020-04-06
[WIP] Add JSDoc recommendations to eslint

mdmower:


mdmower:
Is this a feature that would be appreciated or found cumbersome? (I'm in favor of it)

In its current form, it only warns on JSDoc problems, so linting still passes.

PR #711, opened 2020-04-15
runtime-fetch: Update amp-geo for subdivision matching

mdmower:
ISO 3166-2 code us-ca is now supported by amp-geo. Support reverting
this hotpatch as well. Note: the subdivision part of ISO 3166-2 can be
alphanumeric.


sebastianbenz:
Does it make sense to add a test covering the new behavior?

sebastianbenz:
(merged to put this into the 2.3 release)

PR #718, opened 2020-04-17
runtime-fetch: Fix incorrect geo regex

mdmower:
The regex used to reverse the `{{AMP_ISO_COUNTRY_HOTPATCH}}` patch had two
problems:

- Did not take into account case when country is not found (all spaces)
- Incorrect subdivision pattern

These are the realistic hotpatch examples that can be found in the wild
today:
```
"                            "
"us                          "
"us us-ca                    "
```


mdmower:
Thanks to @jridgewell for pointing out the error: https://github.com/ampproject/amphtml/pull/27100/files#r409916319

PR #722, opened 2020-04-20
Add module/no-module support

sebastianbenz:
This PR adds the experimental module/nomodule runtime loading support.
ESM module loading support can be added via the `experimentEsm=true`
parameter.

### DIFF
--- a/packages/optimizer/lib/transformers/RewriteAmpUrls.js
+++ b/packages/optimizer/lib/transformers/RewriteAmpUrls.js
@@ -68,6 +68,9 @@ class RewriteAmpUrls {
     while (node) {
       if (node.tagName === 'script' && this._usesAmpCacheUrl(node.attribs.src)) {
         node.attribs.src = this._replaceUrl(node.attribs.src, host);
+        if (params.experimentEsm) {
+          this._addEsm(node, node.attribs.src.endsWith('v0.js'));
+        }
         referenceNode = this._addPreload(head, referenceNode, node.attribs.src, 'script');
       } else if (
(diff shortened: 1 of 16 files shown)


sebastianbenz:
FYI: @mdmower @patrickkettner

sebastianbenz:
...and FYI @schlessera @westonruter - we should add this to the wp plugin once the validator supports mjs files.

westonruter:
Filed issue for PHP Optimizer: https://github.com/ampproject/amp-wp/issues/4607

packages/optimizer/lib/transformers/RewriteAmpUrls.js
```diff
@@ -101,6 +107,31 @@ class RewriteAmpUrls {
...
   }
 
+  _addEsm(scriptNode, preload) {
+    const esmScriptUrl = scriptNode.attribs.src.replace(/\.js$/, '.mjs');
+    if (preload) {
+      const preload = createElement('link', {
+        as: 'script',
+        crossorigin: 'anonymous',
+        href: esmScriptUrl,
+        rel: 'preload',
```
patrickkettner, line 117: have we considered using `modulepreload`?

kristoferbaxter: Its unsupported in Safari and Firefox :(

patrickkettner: is they are both included to supporting browsers try to prefetch twice?

sebastianbenz:
Thanks for the review!

Issue #650, opened 2020-03-18
Runtime version getting timeout


c0b41:
my nextjs application using `optimizer` but last 24 hours getting runtime version timeout error, also this happen before 3-4 months ago maybe google cloud outages happen latelty.

so we need some fallback version for runtime-version

https://github.com/ampproject/amp-toolbox/blob/master/packages/runtime-version/lib/RuntimeVersion.js#L76-L86

sebastianbenz:
I haven't heard of any google cloud outages. Did this happen in prod? 

But we definitely should have some kind of fallback. This is also a DevX problem as development is hindered while offline. 

My suggestion would be: 

1. download and cache the latest runtime version + v0.css on install.
2. fallback to cached version on time out (500ms?) 
3. update cached version on every successful download

c0b41:
@sebastianbenz this happen in production.

c0b41:
@sebastianbenz also we can't controll this behavior see https://github.com/zeit/next.js/blob/2ba352da39ee00b6595aecdc9ffb2f103e803a85/packages/next/next-server/server/optimize-amp.ts#L9

sebastianbenz:
Thanks for the details. This was most likely an issue in your prod environment. The runtime version is served from the same endpoint as the AMP runtime which had no outages. Nevertheless, we should fix optimizer falling because of this. 

Sidenote: it's going to be possible to configure optimizer in Next.js once https://github.com/zeit/next.js/pull/10705 has landed in prod.

c0b41:
@sebastianbenz thanks again maybe my server had some issue, either way following next pull request

rajd90:
I'm getting the same issue when using the optimizer to build from my dev machine...

FetchError: request to https://cdn.ampproject.org/rtv/metadata failed, reason: read ECONNRESET
    at ClientRequest.<anonymous> 
    at ClientRequest.emit (events.js:210:5)
    at TLSSocket.socketErrorListener (_http_client.js:406:9)
    at TLSSocket.emit (events.js:210:5)
    at emitErrorNT (internal/streams/destroy.js:92:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:60:3)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  message: 'request to https://cdn.ampproject.org/rtv/metadata failed, reason: read ECONNRESET',
  type: 'system',
  errno: 'ECONNRESET',
  code: 'ECONNRESET'
}
[...]

PR #736, opened 2020-04-28
Extract runtime parameters

sebastianbenz:
This PR extracts fetching and configuration of runtime parameters into a
separate function outside the transformers. This externalizes and
centralizes runtime parameter handling enabling, removing the
responsibility of handling runtime configuration inside transformers
(with the potential for duplication) and enabling future optimizations
of file system based caching of runtime artifacts (see #650).

This PR has a breaking change for the experimental RTV runtime URL re-writing as it now needs to  be enabled explicitly via `rtv: true`.


sebastianbenz:
@mdmower would be great if you could take a look. You probably know most about how runtime version parameters need to be handled. Thanks!

packages/optimizer/lib/fetchRuntimeParameters.js
```diff
@@ -0,0 +1,91 @@
...
+const validatorRules = require('@ampproject/toolbox-validator-rules');
+const {AMP_CACHE_HOST, AMP_RUNTIME_CSS_PATH, appendRuntimeVersion} = require('./AmpConstants.js');
+
+/**
+ * Creates the runtime parameters used by the transformers. Runtime parameters might
+ * change at runtime with new AMP releases.
+ *
+ * @param {Object} config - the AMP Optimizer config
+ * @param {Object} customRuntimeParameters - user defined runtime parameters
+ */
```
mdmower, line 27: You might consider adding a `@returns` to quickly identify the parameters that are typically returned. It's a bit difficult to identify them all by reading through the code. Similarly, in the description, you might mention which parameters are simply validated, which parameters are fetched and verified via network, and the rest pass-through untouched.

packages/optimizer/lib/DomTransformer.js
```diff
@@ -145,13 +144,13 @@ class DomTransformer {
    * @param {Array.<Transformer>} config.transformations - a list of transformers to be applied.
    */
   setConfig(config) {
```
mdmower, line 146: I might opt for `this.config = this.setConfig(config);` in the constructor so that it's easier to identify that DomTransformer has property `config` available without digging into a function.

packages/optimizer/spec/transformers/valid/AmpBoilerplateTransformer/always_inlines_v0css/input.html
```diff
@@ -1,3 +1,9 @@
+<!--
+{
+  "ampRuntimeVersion": "0123456789",
```
mdmower, line 3: Since AMP has decided [RTVs must be 15-digits](https://github.com/ampproject/amphtml/issues/27579), we might want to get in the habit of using valid versions in examples and tests.

packages/optimizer/lib/fetchRuntimeParameters.js
sebastianbenz: Done

packages/optimizer/lib/DomTransformer.js
sebastianbenz: This duplicates initializing the field in the constructor and in the setter. Imo setX implies that there's a field (my Java roots might be showing trough here...)

PR #763, opened 2020-05-11
SSR: Add support for media, sizes and heights attribute

sebastianbenz:
This will make it possible to remove the boilerplate even if the `media`, `sizes` or `heights` attributes are used. These attributes will be transformed into CSS media queries instead.


sebastianbenz:
@westonruter @schlessera PTAL - I haven't yet implemented the CSS size check.

packages/optimizer/lib/transformers/ApplyCommonAttributes.js
```diff
@@ -0,0 +1,254 @@
...
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+const parseSizes = require('../parseSizes');
+const {appendChild, createElement, insertText, hasAttribute} = require('../NodeUtils');
+const ID_PREFIX = 'i-amp-';
```
westonruter, line 20: The prefix `i-amphtml-` is disallowed, but `i-amp-` is OK. It seems somewhat strange why `i-amp-` wasn't chosen rather for the internal prefix since it has fewer bytes.

packages/optimizer/lib/transformers/ApplyCommonAttributes.js
```diff
@@ -0,0 +1,254 @@
...
+      // The user agent will pick a width from the sizes attribute, using the first item with a <media-condition> (the part in parentheses) that evaluates to true.
+      // This means, we have to reverse the order the media queries in CSS to emulate this behavior (the last definition has precedence).
+      heights: heights.values.reverse(),
+    });
+  }
+
+  toString() {
+    let result = '';
+    for (const {heights, defaultHeight, id} of this.heights) {
+      const selector = `#${id}:first-child`;
```
westonruter, line 159: Shouldn't this be:

```suggestion
      const selector = `#${id}>:first-child`;
```

See https://github.com/ampproject/amp-wp/pull/4482#discussion_r421605106

packages/optimizer/lib/transformers/ApplyCommonAttributes.js
```diff
@@ -0,0 +1,254 @@
...
+      return;
+    }
+    if (!customStyles) {
+      customStyles = createElement('style', {
+        'amp-custom': '',
+      });
+      appendChild(head, customStyles);
+      insertText(customStyles, '');
+    }
+    customStyles.children[0].data += styles;
```
westonruter, line 230: As noted in the description, we'll need to make sure this doesn't go above 75000 bytes.

packages/optimizer/lib/transformers/ApplyCommonAttributes.js
```diff
@@ -0,0 +1,254 @@
...
+   * Returns an existing id or generates a new one.
+   *
+   * @param {Node} node
+   */
+  getOrCreateId(node) {
+    if (hasAttribute(node, 'id')) {
+      return node.attribs.id;
+    }
+    node.attribs = node.attribs || [];
+    node.attribs.id = ID_PREFIX + this.counter;
```
westonruter, line 248: Just in case…

```suggestion
    node.attribs.id = ID_PREFIX + this.counter.toString();
```

packages/optimizer/spec/transformers/valid/ServerSideRendering/converts_heights_attribute_to_css/input.html
```diff
@@ -2,32 +2,27 @@
   <head>
     <script async src=https://cdn.ampproject.org/v0.js></script>
     <script async custom-element=amp-img src=https://cdn.ampproject.org/v0/amp-img-0.1.js></script>
+    <noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
     <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;
 -moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes-amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes-amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes-amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes-amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style>
-    <noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
```
schlessera, line 7: Why the change in order here?

packages/optimizer/spec/transformers/valid/ServerSideRendering/converts_media_attribute_to_css/expected_output.html
```diff
@@ -0,0 +1,28 @@
+<html ⚡ i-amphtml-layout i-amphtml-no-boilerplate>
+<head><style amp-runtime></style>
+  <script async src="https://cdn.ampproject.org/v0.js"></script>
+  <script async custom-element="amp-img" src="https://cdn.ampproject.org/v0/amp-img-0.1.js"></script>
+
+
```
schlessera, line 6: Removal of the boilerplate shouldn't leave empty lines behind.

packages/optimizer/lib/parseSizes.js
```diff
@@ -0,0 +1,72 @@
...
+let Size;
+
+/**
+ * Parses the text representation of sizes into Sizes object.
+ * See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Attributes.
+ * See https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-sizes
+ * See https://html.spec.whatwg.org/multipage/images.html#parsing-a-sizes-attribute
+ *
+ * @param {String} string
+ * @returns {!Sizes}
```
schlessera, line 43: Why the "non-nullable" modifier `!` here specifically? Wouldn't that be the default without the modifier anyways? It seems like that distinction isn't use anywhere else.

packages/optimizer/lib/transformers/ApplyCommonAttributes.js
```diff
@@ -0,0 +1,254 @@
...
+ * Transforms a media attribute into CSS by negating the existing media query.
+ */
+class MediaTransformer {
+  constructor() {
+    this.media = new Map();
+  }
+
+  transform(node, id) {
+    // normalize whitespace
+    let mediaString = node.attribs.media.replace(/\s+/g, ' ');
```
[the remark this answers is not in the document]
sebastianbenz: parseSizes is unrelated to the media attribute or am I missing something?

packages/optimizer/lib/parseSizes.js
sebastianbenz: It's a closure compiler flag used by the AMP runtime. Removed it as we don't need it.

packages/optimizer/lib/transformers/ApplyCommonAttributes.js
```diff
@@ -0,0 +1,254 @@
...
+      }
+    }
+  }
+
+  /**
+   * Returns an existing id or generates a new one.
+   *
+   * @param {Node} node
+   */
+  getOrCreateId(node) {
```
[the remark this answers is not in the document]
sebastianbenz: True. This is not straightforward to implement in our case unfortunately.

packages/optimizer/lib/transformers/ApplyCommonAttributes.js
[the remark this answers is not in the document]
schlessera: Oh, nvm, I was thinking of the media conditions that are part of the `sizes` attribute.

packages/optimizer/spec/transformers/valid/ServerSideRendering/converts_media_attribute_to_css/expected_output.html
sebastianbenz: that's going to be removed by a different transformer (not used in the test)

packages/optimizer/spec/transformers/valid/ServerSideRendering/converts_heights_attribute_to_css/input.html
sebastianbenz: reverted

packages/optimizer/lib/transformers/ApplyCommonAttributes.js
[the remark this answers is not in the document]
sebastianbenz: I'll ignore this case for now. Re-running the transformer would simply re-use the existing id. Pre-existing Ids should not be avoided by our prefix.

sebastianbenz:
I've addressed the comments. In regards to checking for the CSS limit. I'm currently lending towards not including this check:

* CSS optimization would need to be run twice (before, to be able to get the correct byte count and after, to optimize the generated CSS rules).
* I'd prefer to fail on not being able to remove the boilerplate, the perf implications of not being able to remove the boilerplate because of this are a lot harder to detect otherwise.

sebastianbenz:
@westonruter @schlessera PTAL

westonruter:
>  In regards to checking for the CSS limit. I'm currently lending towards not including this check

Doesn't this run the risk of the Optimizer generating invalid AMP from valid AMP input? If the CSS is close to the limit and then optimization causes it to go over, that seems not ideal?

packages/optimizer/spec/transformers/valid/ServerSideRendering/converts_media_attribute_to_css/expected_output.html
sebastianbenz: This will be fixed by a different transformer

packages/optimizer/lib/transformers/ApplyCommonAttributes.js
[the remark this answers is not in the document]
sebastianbenz: Done

sebastianbenz:
Merging as there's no further feedback that needs to be addressed.

schlessera:
@sebastianbenz There seems to be an issue with the CSS in here that was raised twice but always resolved without a comment, so I assume that GitHub resolved it because of an unrelated change: https://github.com/ampproject/amp-toolbox/pull/763#discussion_r425816316

Or am I missing something maybe?

sebastianbenz:
Hah! Not sure how this got lost. Fixed in #811

sebastianbenz:
Thanks for picking up on this!

PR #771, opened 2020-05-13
Fixes postinstall for optimizer on windows

andreban:
- The `postinstall` script runs `scripts/init.js` via a shebang.
  This doesn't work on Windows, so we change the script to execute
  `node scripts/init.js` explicitly.
- Checked and the `postinstall` hook still works on Unix (via WSL).
- May be worth test on MacOS (expect it to work though)

PS: Had to push with `-no-verify` as the build script breaks on Windows.

### DIFF
--- a/packages/optimizer/package.json
+++ b/packages/optimizer/package.json
@@ -9,7 +9,7 @@
     "directory": "packages/optimizer"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


andreban:
Already fixed via #770

sebastianbenz:
Thanks!!! Closing as duplicate of #770

PR #772, opened 2020-05-13
add srcset transformer

sebastianbenz:


packages/optimizer/lib/transformers/OptimizeImages.js
```diff
@@ -0,0 +1,249 @@
...
+ *    pointing to a version of the `src` image with the given `width`. If no image is available, it should
+ *    return a falsy value. For example:
+ *
+ *       (src, width) => `${src}?width=${width}`
+ */
+class OptimizeImages {
+  constructor(config) {
+    this.log = config.log;
+    this.enabled = !!config.optimizeImages;
+    this.imageOptimizer = config.imageOptimizer || DEFAULT_IMAGE_OPTIMIZER;
```
sebastianbenz, line 158: I've also considered to always require users to provide a function and get rid of the optimizeImages flag. Not sure if the default is very useful.

packages/optimizer/lib/transformers/OptimizeImages.js
```diff
@@ -0,0 +1,249 @@
...
+'use strict';
+
+const {hasAttribute, nextNode, firstChildByTag} = require('../NodeUtils');
+const {skipNodeAndChildren} = require('../HtmlDomHelper');
+const {isValidImageSrcURL} = require('../URLUtils');
+
+const MAX_IMG_SIZE = 820;
+const MIN_WIDTH_TO_ADD_SRCSET_IN_RESPONSIVE_LAYOUT = 300;
+const NUM_SRCSET_DPR = [1.0, 2.0, 3.0];
+const SRCSET_WIDTH = [
```
patrickkettner, line 26: whole lot of magic numbers, does it make sense to add a comment as to why these numbers are being used?

packages/optimizer/lib/transformers/OptimizeImages.js
```diff
@@ -0,0 +1,249 @@
...
+/**
+ * Default implementation that does not perform any kind of image optimization, but generates a `srcset` string
+ * by appending the width to the file name (e.g. image.jpg => image.32w.jpg).
+ *
+ * @param {string} src - the image src
+ * @param {number} width - the required widths (in px)
+ * @returns {string|undefined} - the image URL or undefined if no image is available in this dimension
+
+ */
+const DEFAULT_IMAGE_OPTIMIZER = (src, width) => {
```
patrickkettner, line 57: is this private? theres a lot of assumptions below (src exists, and is a string, with exists, etc)

packages/optimizer/lib/transformers/OptimizeImages.js
```diff
@@ -0,0 +1,249 @@
...
+ */
+class SrcsetWidth {
+  constructor() {
+    this.widthList_ = [];
+  }
+
+  /**
+   * Sets the base width, i.e., renderered dimension measured in CSS pixels.
+   * Returns true if srcset is needed, that is, we'll resize the image to at
+   * least 2 legitimate widths.
```
patrickkettner, line 79: what is a legitimate width?

packages/optimizer/lib/transformers/OptimizeImages.js
```diff
@@ -0,0 +1,249 @@
...
+      if (width != previousWidth) {
+        this.widthList_.push(width);
+      }
+
+      previousWidth = width;
+    }
+  }
+
+  /**
+   *  Returns true if there is more legitimate width.
```
patrickkettner, line 110: im not sure if this means "returns true is there are additional widths that are also legitimate", or "returns true is there is a different width that is preferable to the current width"

packages/optimizer/lib/transformers/OptimizeImages.js
```diff
@@ -0,0 +1,249 @@
...
+    return this.widthList_.length > 0;
+  }
+
+  /**
+   * Returns the current legitimate width and moves the state to the next one.
+   */
+  nextWidth() {
+    const nextWidth = this.widthList_[this.widthList_.length - 1];
+    this.widthList_.pop();
+    return nextWidth;
```
patrickkettner, line 122: any reason this wouldn't just be something like
```js
return this.widthList_.pop()
```

or is the variable allocation for readability?

packages/optimizer/lib/transformers/OptimizeImages.js
patrickkettner: why is that?
My preference would be to either keep it or have a few examples documented somewheres to ease the onboarding

packages/optimizer/lib/transformers/OptimizeImages.js
```diff
@@ -0,0 +1,249 @@
...
+      return;
+    }
+    const src = imageNode.attribs.src;
+    // Check if it's a relative path or a valid http(s) URL.
+    if (!isValidImageSrcURL(src)) {
+      return;
+    }
+
+    // No srcset is added if the image ends with a `,` (comma). See
+    // http://b/127535381 for context.
```
patrickkettner, line 198: is it ok reference corplinks on a public repo?

packages/optimizer/lib/transformers/OptimizeImages.js
```diff
@@ -0,0 +1,249 @@
...
+    if (src.endsWith(',')) {
+      return;
+    }
+    const width = imageNode.attribs.width;
+
+    // TODO(b/113271759): Handle width values that include 'px' (probably others).
```
patrickkettner, line 204: same as above

packages/optimizer/lib/transformers/OptimizeImages.js
sebastianbenz: My plan was to go with samples as well. Current thinking is to provide implementations for:

* cloudiary
* thumbor
* static site (via eleventy-img wrapper)
* custom sample

If we do this, then it probably makes sense to remove the default.

packages/optimizer/lib/transformers/OptimizeImages.js
sebastianbenz: will turn this into a sample, it's not useful this way

packages/optimizer/lib/transformers/OptimizeImages.js
sebastianbenz: yeah, it's not nice, but better than leaving it completely undocumented. It's a common pattern in the AMP repo.

packages/optimizer/lib/transformers/OptimizeImages.js
sebastianbenz: added comments

packages/optimizer/lib/transformers/OptimizeImages.js
sebastianbenz: added comment to the list of supported srcset widths

packages/optimizer/lib/transformers/OptimizeImages.js
sebastianbenz: no reason...

sebastianbenz:
@patrickkettner PTAL

packages/optimizer/lib/transformers/OptimizeImages.js
patrickkettner: what would downstream tools (WP, next, etc) do? provide their own functions?

sebastianbenz: Correct, this is most likely something that users need to provide as frameworks usually don't optimize images by default.

packages/optimizer/lib/transformers/OptimizeImages.js
sebastianbenz: rephrased

packages/optimizer/lib/transformers/OptimizeImages.js
```diff
@@ -0,0 +1,217 @@
...
+ * limitations under the License.
+ */
+
+'use strict';
+
+const {hasAttribute, nextNode, firstChildByTag} = require('../NodeUtils');
+const {skipNodeAndChildren} = require('../HtmlDomHelper');
+const {isValidImageSrcURL} = require('../URLUtils');
+
+// Don't generate srcsets for images larger than MAX_IMG_SIZE
```
patrickkettner, line 23: any reason this is the max that is worth documenting? based on stats? device? processing power?

packages/optimizer/lib/transformers/OptimizeImages.js
```diff
@@ -0,0 +1,217 @@
...
+const MAX_IMG_SIZE = 1500;
+
+// Don't generate srcset's for images with width smaller than MIN_WIDTH_TO_ADD_SRCSET_IN_RESPONSIVE_LAYOUT
+// this avoids generating srcsets for images with a responsive layout where width/height define the aspect ration.
+const MIN_WIDTH_TO_ADD_SRCSET_IN_RESPONSIVE_LAYOUT = 100;
+
+// All legimate srcset widths.
+const SRCSET_WIDTH = [39, 56, 82, 100, 150, 300, 500, 750, 1000, 1500, 2000, 2500];
+
+// The maximum number of values. We'll take the initial image width and generate more width values by
```
patrickkettner, line 33: does "values" in this context mean "maximum number of [`srcset` sources]"?

packages/optimizer/lib/transformers/OptimizeImages.js
```diff
@@ -0,0 +1,217 @@
...
+ * - `imageOptimizer`: a function for customizing the srcset generation. The function should return a URL
+ *    pointing to a version of the `src` image with the given `width`. If no image is available, it should
+ *    return a falsy value. For example: (src, width) => `${src}?width=${width}`.
+ */
+class OptimizeImages {
+  constructor(config) {
+    this.log = config.log;
+    this.imageOptimizer = config.imageOptimizer;
+    this.srcsetWidth = new SrcsetWidth();
+    // TODO turn these into options
```
patrickkettner, line 125: always cool to link to a tracking bug

packages/optimizer/lib/transformers/OptimizeImages.js
sebastianbenz: Good point. Leftover from cache implementation. Changed it to the max supported value.

packages/optimizer/lib/transformers/OptimizeImages.js
sebastianbenz: yes, rephrased

sebastianbenz:
Thanks a lot for the great review @patrickkettner !

PR #780, opened 2020-05-15
Condense linter output for image alt text check

Dbrown910:
@sebastianbenz


packages/linter/src/rules/ImagesHaveAltText.ts
```diff
@@ -3,16 +3,31 @@ import { Rule } from "../rule";
...
+      output +=
+        el
+          .replace('{"', "")
+          .replace('":', " (found ")
+          .replace('"', "")
+          .replace("}", "") + " times)\n";
+    });
+
+    return Object.keys(imgsWithoutAlt).length > 0
+      ? this.warn("Missing alt text from images: \n" + output)
```
sebastianbenz, line 30: Shouldn't we fail?

packages/linter/src/rules/ImagesHaveAltText.ts
```diff
@@ -3,16 +3,31 @@ import { Rule } from "../rule";
...
+        } else {
+          imgsWithoutAlt[elem.attribs.src] += 1;
+        }
       }
     });
 
-    return imgsWithoutAlt.length > 0
-      ? this.warn(`Missing alt text from images: \n` + imgsWithoutAlt)
+    let output = "";
+    let m = `${JSON.stringify(imgsWithoutAlt)}`.split(",");
```
sebastianbenz, line 19: `imgsWithoutAlt.map(img => ....).join('\n')` would be a more idiomatic approach.

packages/linter/src/rules/ImagesHaveAltText.ts
Dbrown910: I wanted to. But according to the feedback I've gotten it should be a warning since the Story could still be valid AMP and surfaced. Its mostly to stay aligned with the official AMP test tool.

sebastianbenz:
Please run `npm run lint:fix`

PR #781, opened 2020-05-15
Fix lts

sebastianbenz:
The new runtime parameter handling resulted in the LTS option no longer
working. Moving lts + ampRuntimeVersion in fetchRuntimeParameters.

This PR also improves logging and fixes a bug where the styles fallback
did not work.


sebastianbenz:
@mdmower would be great if you could take a look.

packages/optimizer/lib/RuntimeHostHelper.js
```diff
@@ -17,11 +17,12 @@
 
 const {AMP_CACHE_HOST, appendRuntimeVersion} = require('./AmpConstants.js');
 
-function calculateHost({ampUrlPrefix = AMP_CACHE_HOST, ampRuntimeVersion, lts, rtv = false}) {
-  if (lts && ampRuntimeVersion) {
-    throw new Error('lts flag is not compatible with runtime version parameter');
```
mdmower, line 22: I think this error is worth keeping, but `if (lts && ampRuntimeVersion)` should be replaced with `if (lts && rtv)`.

packages/optimizer/lib/fetchRuntimeParameters.js
```diff
@@ -149,11 +155,17 @@ async function fetchAmpRuntimeVersion_(context) {
...
     version: await config.runtimeVersion.currentVersion({ampUrlPrefix, lts}),
     maxAge: MaxAge.create(AMP_RUNTIME_MAX_AGE).toObject(),
   };
-  if (versionKey) {
+  if (!ampRuntimeData.version && ampUrlPrefix !== AMP_CACHE_HOST) {
+    config.log.error(
+      `Could not download runtime version from ${ampUrlPrefix}. Falling back to ${AMP_CACHE_HOST}`
+    );
+    ampRuntimeData = await fetchLatestRuntimeData_({config, AMP_CACHE_HOST, lts}, versionKey);
+  } else if (versionKey) {
```
mdmower, line 168: Do you want to store bad results? If not:
`} else if (ampRuntimeData.version && versionKey) {`

packages/optimizer/lib/RuntimeHostHelper.js
sebastianbenz: Good point. Added.

packages/optimizer/lib/fetchRuntimeParameters.js
sebastianbenz: Done

PR #798, opened 2020-05-27
fix: add missing dependencies

merceyz:
**What's the problem this PR addresses?**

`@ampproject/toolbox-core` and `@ampproject/toolbox-optimizer` is using undeclared dependencies which breaks Yarn PnP support in Next.js (https://github.com/yarnpkg/berry/runs/711731982?check_suite_focus=true#step:5:51).

x-ref: https://github.com/zeit/next.js/pull/13354

**How did you fix it?**

- Added `lru-cache` to `@ampproject/toolbox-core`
- Added `cross-fetch`, `dom-serializer`, `postcss` to `@ampproject/toolbox-optimizer`


sebastianbenz:
Thanks for the fix! This required a bit more work as I needed to migrate to the new dom-serializer API. Closing this in favor of #799

PR #799, opened 2020-05-27
[optimizer] add missing runtime dependencies

sebastianbenz:


sebastianbenz:
CLI has been signed if #798

merceyz:
Seems you removed `cross-fetch` which is also needed

sebastianbenz:
@merceyz cross-fetch is a transitive dependency of toolbox-core

merceyz:
That makes us rely on hoisting which is a really, really bad idea, if you want to require it from toolbox-core you should use createRequire/have toolbox-core export it.

sebastianbenz:
Thanks for the explanation. Will add it.

merceyz:
Awesome, thank you. Also, the same issue is present for toolbox-cli, it tries to use toolbox-core without declaring it

PR #805, opened 2020-05-28
Support media attribute for hero image preloading

sebastianbenz:


packages/optimizer/lib/AmpConstants.js
```diff
@@ -15,10 +15,25 @@
...
   AMP_FORMATS: ['AMP', 'AMP4EMAIL', 'AMP4ADS'],
   AMP_RUNTIME_CSS_PATH: '/v0.css',
   appendRuntimeVersion: (prefix, version) => prefix + '/rtv/' + version,
+  isTemplate: (node) => {
+    if (node.tagName === 'template') {
+      return true;
+    }
+    if (
+      node.tagName === 'script' &&
+      hasAttribute(node, 'template') &&
```
patrickkettner, line 32: is there a reason to check for its existence? if it doesn't, the next line would just be against undefined, right?

seems like it could be terse-d up into 

```js
return node.tagName === 'template' || node.tageName === 'script' && node.attribs.template === 'amp-mustache'
```

or even 
```js
return node.tagName === 'template' || node.attribs.template === 'amp-mustache'
```
no worries either way

patrickkettner:
lgtm

packages/optimizer/lib/AmpConstants.js
sebastianbenz: attribs might be undefined

PR #806, opened 2020-05-28
Sort linter output

Dbrown910:
@sebastianbenz


packages/linter/src/cli.ts
```diff
@@ -53,6 +53,7 @@ export function cli(argv: string[], logger = console, cmd = "amplint") {
       /^(googlebot_desktop|googlebot_mobile|chrome_desktop|chrome_mobile)$/i,
       "googlebot_mobile"
     )
+    .option(`-n, --no-supress`, "supress passing tests in output")
```
sebastianbenz, line 56: The description seems to contradict the parameter name (supress vs no supress).

packages/cli/lib/cmds/lint.js
```diff
@@ -19,7 +19,7 @@
 const {cli} = require('@ampproject/toolbox-linter');
 
 function lint(args, logger) {
-  return cli(['dummy'].concat(args), logger, 'amp lint'); // "dummy" to simulate process.argv
+  return cli(['dummy'].concat(args).concat(['-n']), logger, 'amp lint'); // "dummy" to simulate process.argv
```
sebastianbenz, line 22: Am I right that this will by default supress the passing tests? How do I print the passing tests? Do I need to set `-n=false`? I'd find it to be more intuitive to name the attribute something like `--show-passing` which is disabled by default.

packages/linter/src/cli.ts
Dbrown910: whoops typo. I did change this but forgot to commit the edit. Will fix

packages/cli/lib/cmds/lint.js
Dbrown910: `-n` will unsupress the passing tests since they are hidden by default. I had to add this flag here to pass the toolbox testing suite, as one of its checks is "has at least one passing test". 
I can change the flag to `-s, --show-passing`

PR #807, opened 2020-05-28
List all supported options in README

sebastianbenz:


packages/optimizer/README.md
```diff
@@ -48,17 +50,214 @@ ampOptimizer.transformHtml(originalHtml).then((optimizedHtml) => {
...
+* [preloadHeroImage](#preloadheroimage)
+* [verbose](#verbose)
+
+#### `autoAddBoilerplate`
 
-It's possible to pass incomplete documents and AMP Optimizer will add any 
-missing tags and extension imports required by a valid AMP document. 
+Automatically inject any missing markup required by AMP.
+
+* name: `autoAddBoilerplate`
```
patrickkettner, line 96: this isn't AMP Boilerplate, right? naming is confusing if I am correct (since you advocate for it being removed laster on)

sebastianbenz: Yup, but changing this would be a breaking API change :-(. Anyway, renamed it to autoAddMandatoryTags and kept the old name as well.

PR #810, opened 2020-05-29
Fix id generation

sebastianbenz:


packages/optimizer/lib/transformers/ApplyCommonAttributes.js
```diff
@@ -216,8 +219,11 @@ class ApplyCommonAttributes {
         if (hasAttribute(node, attribute)) {
           try {
             const id = this.getOrCreateId(node);
-            transformer.transform(node, id);
+            const nodeHasBeenTransformed = transformer.transform(node, id);
             this.transformedNodes.push(node);
+            if (nodeHasBeenTransformed && !node.attribs.id) {
+              node.attribs.id = id;
+            }
```
schlessera, line 226: Right now, the ID is wasted in this case, and the counter just moves on. This might result in weird markup where you have three elements named `i-amp-3`, `i-amp-7` & `i-amp-19`, which is not clean and quite misleading. It could also make the tests cumbersome, as negative cases will have an unnecessary impact on the positive cases.

I think the simplest solution with your current code is to just decrease the counter again if the ID was not used.
```suggestion
            if (nodeHasBeenTransformed && !node.attribs.id) {
              node.attribs.id = id;
            } else {
              // The ID was not used, so we decrease the internal counter again to
              // resurface it on the next element that needs an ID.
              this.counter--;
            }
```

sebastianbenz: done

schlessera:
LGTM!

sebastianbenz:
Thanks!

PR #814, opened 2020-05-29
Move hero image preload out of experimental

sebastianbenz:
This also changes the transformer to only run if there a no pre-existing image preloads.

### DIFF
--- a/packages/optimizer/lib/transformers/PreloadHeroImage.js
+++ b/packages/optimizer/lib/transformers/PreloadHeroImage.js
@@ -36,7 +36,7 @@ const TINY_IMG_THRESHOLD = 150;
 class PreloadHeroImage {
   constructor(config) {
     this.log = config.log;
-    this.enabled = config.preloadHeroImage !== false && config.experimentPreloadHeroImage;
+    this.enabled = config.preloadHeroImage !== false;
   }
   async transform(root, params) {
     if (!this.enabled || params.preloadHeroImage === false) {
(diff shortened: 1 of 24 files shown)


packages/optimizer/lib/transformers/PreloadHeroImage.js
```diff
@@ -78,6 +83,28 @@ class PreloadHeroImage {
...
   }
 
+  hasExistingImagePreload(head) {
+    for (const node of head.children) {
+      if (this.isImagePreload(node)) {
+        return true;
+      }
+    }
+    return false;
+  }
```
patrickkettner, line 93: continuing my series of "unhelpful comments that make the code less readable"...

```js
return [...head.children].any(this.isImagePreload)
```

packages/optimizer/spec/end-to-end/body-only/expected_output.default.html
```diff
@@ -6,6 +6,7 @@
   <link rel="preload" href="https://cdn.ampproject.org/v0.js" as="script">
   <script data-auto async src="https://cdn.ampproject.org/v0.js"></script>
   <script async src="https://cdn.ampproject.org/v0/amp-video-0.1.js" custom-element="amp-video"></script>
+  <link rel="preload" href="https://amp.dev/static/samples/img/tokyo.jpg" as="image" data-hero>
```
patrickkettner, line 9: just for my own edification - what is the purpose of the data attr?

packages/optimizer/lib/transformers/PreloadHeroImage.js
sebastianbenz: Why the destructuring? I think `head.children.some(this.isImagePreload);` does the trick as well (and is easy to read).

packages/optimizer/spec/end-to-end/body-only/expected_output.default.html
sebastianbenz: For being able to track which images have been identified as hero images

westonruter:
For images that have `srcset` and `sizes`, this may not work as expected: #823.

PR #825, opened 2020-06-05
Added a check for og:image in the name attr

Dbrown910:
@sebastianbenz


sebastianbenz:
Can you please add a test for this

PR #828, opened 2020-06-08
Do not preload hero images using srcset

### LINKED ISSUE #823: PreloadHeroImage should use imagesrcset attribute not srcset
I just noticed that the `PreloadHeroImage` transformer is adding a preload `link` with a `srcset` attribute when it should rather be adding `imagesrcset` per https://web.dev/preload-responsive-images/

https://github.com/ampproject/amp-toolbox/blob/4b59cdceddcfe3bb4f39c7e82908561e3576e87a/packages/optimizer/lib/transformers/PreloadHeroImage.js#L62-L81

Also, should not an `imagesizes` attribute be added to the `link` if the `img` has `sizes` attribute?

Lastly, it seems that the Validator hasn't been updated to allow `srcset` (or `imagesrcset`) or `sizes` (or `imagesizes`) on the `link` element:

https://github.com/ampproject/amphtml/blob/1b84e3c0d9899d826c32bdccd99597409ccf0e1d/validator/validator-main.protoascii#L287-L304

Doesn't the transformer currently thus create invalid AMP?

sebastianbenz:
Fixes #823

### DIFF
--- a/packages/optimizer/lib/transformers/PreloadHeroImage.js
+++ b/packages/optimizer/lib/transformers/PreloadHeroImage.js
@@ -69,17 +69,9 @@ class PreloadHeroImage {
       preload.attribs.media = heroImage.media;
     }
     if (heroImage.srcset) {
-      try {
-        parseSrcSet(heroImage.srcset);
-        preload.attribs.srcset = heroImage.srcset;
-      } catch (err) {
-        this.log.warn(
-          `Invalid srcset '${heroImage.srcset}' for amp-img src '${heroImage.src}'`,
-          err.message
-        );
-      }
+      this.log.debug("Could not preload hero image as it's using srcset", heroImage.src);
+      return;
     }
-    this.log.debug('Preloading hero image: ', heroImage.src);
     insertAfter(head, preload, referenceNode);
   }
 
--- a/packages/optimizer/lib/transformers/PreloadHeroImage.js
+++ b/packages/optimizer/lib/transformers/PreloadHeroImage.js
@@ -69,7 +69,10 @@ class PreloadHeroImage {
       preload.attribs.media = heroImage.media;
     }
     if (heroImage.srcset) {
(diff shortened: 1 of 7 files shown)


packages/optimizer/lib/transformers/PreloadHeroImage.js
```diff
@@ -69,17 +69,9 @@ class PreloadHeroImage {
...
-      try {
-        parseSrcSet(heroImage.srcset);
-        preload.attribs.srcset = heroImage.srcset;
-      } catch (err) {
-        this.log.warn(
-          `Invalid srcset '${heroImage.srcset}' for amp-img src '${heroImage.src}'`,
-          err.message
-        );
-      }
+      this.log.debug("Could not preload hero image as it's using srcset", heroImage.src);
```
westonruter, line 72:
```suggestion
      this.log.debug("Could not preload hero image as it's using srcset, which is currently only supported Chromium-based browsers (see https://web.dev/preload-responsive-images/).", heroImage.src);
```

packages/optimizer/lib/transformers/PreloadHeroImage.js
```diff
@@ -69,17 +69,9 @@ class PreloadHeroImage {
       preload.attribs.media = heroImage.media;
     }
     if (heroImage.srcset) {
```
westonruter, line 71: I wonder if this should be configurable? Should there be a flag to force inclusion of images that have `srcset`?

sebastianbenz: That would cause double image downloads which is something I'd prefer to avoid.

westonruter: But only on non-Chromium browsers. It's not needed now, but it would be useful in the near future to evaluate the performance impact for pages that use `srcset`.

sebastianbenz:
thanks!

PR #835, opened 2020-06-10
More robust cache warmup 

sebastianbenz:
* timeout fetch after 2s to avoid overly long timeouts
* explain how to configure a custom proxy
* fix potential endless loop if fetch fails

Fixes #832

PR #838, opened 2020-06-12
fix(optimizer): add missing dependency node-fetch

merceyz:
**What's the problem this PR addresses?**

In https://github.com/ampproject/amp-toolbox/pull/835 `@ampproject/toolbox-optimizer` started using `node-fetch` but didn't declare it as a dependency.

**How did you fix it?**

Add missing dependency `node-fetch`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -84,8 +84,8 @@
         "@ampproject/toolbox-cache-list": "^2.5.1",
         "@ampproject/toolbox-cache-url": "^2.3.0",
         "@ampproject/toolbox-core": "^2.5.1",
-        "@ampproject/toolbox-linter": "^2.5.1",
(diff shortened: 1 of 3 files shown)


sebastianbenz:
Sorry about missing this again. Is there a simple way to check this via linter?

merceyz:
> Sorry about missing this again

No worries :+1: 

> Is there a simple way to check this via linter?

Yes, there is a ESLint plugin that can handle this.
https://github.com/benmosher/eslint-plugin-import
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md

sebastianbenz:
Thanks! Unfortunately, it looks like it doesn't play well with how lerna
handles dependencies.

>

PR #846, opened 2020-06-22
Fix: MinifyHtml cannot be disabled

### LINKED ISSUE #845: MinifyHtml Transformer cannot be set to false
https://github.com/ampproject/amp-toolbox/blob/bf9cdb032c1c35e1c3a11ed288c081ef7631f239/packages/optimizer/lib/transformers/MinifyHtml.js#L41

Seems `opts.minify` will always result to true, regardless of what is passed into config variable.

sebastianbenz:
Fixes #845


patrickkettner:
LGTM, as long as we are good with `0`, `null`, `undefined` being true

sebastianbenz:
Yup, default is enabled

PR #852, opened 2020-07-01
Improve preload handling with runtime self hosting

sebastianbenz:
* remove v0.js preloads if esm mode is enabled (the esm modules should
be preloaded instead)
* rename existing v0.js preloads

### DIFF
--- a/packages/optimizer/lib/transformers/RewriteAmpUrls.js
+++ b/packages/optimizer/lib/transformers/RewriteAmpUrls.js
@@ -15,7 +15,7 @@
  */
 'use strict';
 
-const {createElement, firstChildByTag, insertAfter, insertBefore} = require('../NodeUtils');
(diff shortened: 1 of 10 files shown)


packages/optimizer/lib/transformers/RewriteAmpUrls.js
```diff
@@ -67,25 +67,46 @@ class RewriteAmpUrls {
...
+        } else {
+          node.attribs.href = this._replaceUrl(node.attribs.href, host);
+        }
       }
       node = node.nextSibling;
     }
 
+    // process preloads later to avoid accidently rewriting the URL
+    for (const preload of preloads) {
+      if (preload) {
```
patrickkettner, line 105: when would this be false?

sebastianbenz: Some scripts should not have a preload. In that case rewrite URL returns `null`.

sebastianbenz:
Thanks for the review

PR #855, opened 2020-07-08
Only inject runtime-host meta tag if host is given

sebastianbenz:
* Only include host in runtime-host meta tag
* Print warning if ampUrlPrefix is not absolute
* Don't inject runtime-host meta tag if ampUrlPrefix is not absolute.
This will fallback to cdn.ampproject.org.

### DIFF
--- a/packages/optimizer/lib/transformers/RewriteAmpUrls.js
+++ b/packages/optimizer/lib/transformers/RewriteAmpUrls.js
@@ -57,6 +57,7 @@ const {calculateHost} = require('../RuntimeHostHelper');
 class RewriteAmpUrls {
   constructor(config) {
     this.esmModulesEnabled = config.experimentEsm;
+    this.log = config.log;
   }
   transform(root, params) {
     const html = firstChildByTag(root, 'html');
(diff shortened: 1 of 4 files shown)


packages/optimizer/lib/transformers/RewriteAmpUrls.js
```diff
@@ -109,13 +110,12 @@ class RewriteAmpUrls {
...
     // runtime-host and amp-geo-api meta tags should appear before the first script
     if (!this._usesAmpCacheUrl(host) && !params.lts) {
-      let versionlessHost;
-      if (this.isAbsoluteUrl_(params.ampUrlPrefix)) {
-        versionlessHost = calculateHost({ampUrlPrefix: params.ampUrlPrefix});
-      } else {
-        versionlessHost = '.';
+      try {
+        const url = new URL(host);
+        this._addMeta(head, 'runtime-host', url.origin);
```
mdmower, line 115: This introduced a regression. It used to be possible to specify a host like `https://example.com/amp-runtime`. This strips away `/amp-runtime`, resulting in incorrect URLs for dynamically loaded extensions.

PR #857, opened 2020-07-08
Proxy support for amp-optimizer

todoa2c:
Proxy support using node-fetch-with-proxy instead of node-fetch.
If you need to use proxy, just set the environment variable `https_proxy`.

Thanks.


todoa2c:
@googlebot I signed it!

sebastianbenz:
Thanks for creating this! One suggestion: as the code for integrating this is trivial, would you mind implementing this directly instead of adding an additional dependency?

sebastianbenz:
Sorry for not being more clear. Integrating `https://www.npmjs.com/package/http-proxy-agent` is fine as it seems to be the "standard" node way of integrating of proxy. What I'd like to avoid is integrating `node-fetch-with-proxy` as it's a super thin wrapper with only a few LOC and super low usage.

todoa2c:
@sebastianbenz Sorry for late reaction.
This time `https-proxy-agent` matches so I added and fixed code.

todoa2c:
@sebastianbenz I checked `cross-fetch` library but it doesn't support agent like `node-fetch`.

Should I replace cross-fetch to node-fetch? or any ideas?

sebastianbenz:
Hm - that's annoying. I think it's fine to replace cross-fetch in that case and we don't release a browser bundle yet which would benefit from cross fetch.

todoa2c:
@sebastianbenz Ok, so let's try replacing "cross-fetch" with "node-fetch" once in a separate branch.
Is it possible to merge this pull request? I'm using next.js now, but I'm having trouble with next.js because it relies on amp-toolbox and I'm having to wait a huge amount of time for postinstall in a proxy environment.

PR #859, opened 2020-07-09
add experimental amp-img -> img transformation

sebastianbenz:


packages/optimizer/lib/transformers/ApplyLayout.js
```diff
@@ -110,8 +110,7 @@ function maybeAddSizerInto(node, layout, width, height) {
     sizer = createIntrinsicSizer(width, height);
   }
   if (sizer) {
-    const referenceNode = node.firstChild;
-    insertBefore(node, sizer, referenceNode);
+    appendChild(node, sizer);
```
sebastianbenz, line 113: Looks like the sizer needs to be added after the img.

packages/optimizer/lib/transformers/PreloadHeroImage.js
```diff
@@ -216,6 +225,35 @@ class PreloadHeroImage {
     }
     return {width: 0, height: 0};
   }
+
+  generateImg(node) {
+    if (!this.experimentImg) {
+      return;
+    }
+    const imgNode = createElement('img', {
+      class: 'class=i-amphtml-fill-content i-amphtml-replaced-content',
```
jridgewell, line 234:
```suggestion
      class: 'i-amphtml-fill-content i-amphtml-replaced-content',
```

packages/optimizer/lib/transformers/PreloadHeroImage.js
```diff
@@ -216,6 +225,35 @@ class PreloadHeroImage {
     }
     return {width: 0, height: 0};
   }
+
+  generateImg(node) {
+    if (!this.experimentImg) {
+      return;
+    }
```
jridgewell, line 232: I think we were only going to do this transform for `amp-img`s that have a `data-hero` attr set. Since we can't guarantee `loading=lazy` behavior for most browsers yet, we don't want to load a bunch of far-from-viewport images.

sebastianbenz: Introduce a max upper bound of 2 hero images and switched to `loading=eager` for these.

packages/optimizer/lib/transformers/PreloadHeroImage.js
sebastianbenz: fixed

packages/optimizer/lib/transformers/ApplyLayout.js
sebastianbenz: Hm. Not sure what I tested there. All good with sizer first.

sebastianbenz:
Adressed the comments, also:

* introduce data-hero attribute which will disable auto hero detection
* don't bail out if there are any kind of image preloads, instead check
the img src to avoid duplicates
* use loading=eager instead of loading=lazy

PTAL

packages/optimizer/lib/transformers/PreloadHeroImage.js
```diff
@@ -216,6 +291,38 @@ class PreloadHeroImage {
...
+    if (!node) {
+      return;
+    }
+    const imgNode = createElement('img', {
+      class: 'i-amphtml-fill-content i-amphtml-replaced-content',
+      decoding: 'async',
+      loading: 'eager',
```
jridgewell, line 305: I think we can drop this attr.

sebastianbenz: done

sebastianbenz:
@jridgewell Thanks for the review!

Issue #862, opened 2020-07-10
ITI: Asset URL rewriting in AMP Optimizer


jonluca:
Moving the conversation here based on https://github.com/ampproject/amppackager/issues/449

We’d like to move forward with adding AMP Packager compatibility mode to the AMP optimizer.

From the previous issue:

> The important thing to note is that in the near future, transformers won't need to produce byte compatible AMPHTML to the AMPPackager. This will greatly simplify things and means that transformers only need to produce valid AMP. So you will be able to use AMP Optimizer out of the box. However, for performance reasons, asset URL rewriting to cdn.ampproject.org is critical in the SXG case and should be added to AMP Optimizer. This could be enabled via a flag as an "AMPPacker compatibility mode".

We feel comfortable in adding in a new flag ( `--amp-packager-mode`?). The base URL would default to `cdn.ampproject.org`, but could be changed to support different AMP caches (is this what the `AMP-Cache-Transform` request header is for?).

sebastianbenz:
Sounds good! I'd prefer to name the flag more generic though: `rewriteUrlsToAmpCache=cdn.ampproject.org`?

One note: rewriting script import URLs is already supported via the [ampUrlPrefix flag](https://github.com/ampproject/amp-toolbox/blob/master/packages/optimizer/lib/transformers/RewriteAmpUrls.js#L35). If the packager flag is set this should be overridden (and print a warning if it is defined as well).

jonluca:
Is there a reason we shouldn't just use the `ampUrlPrefix` flag? It seems like the behavior is going to be the same. Is `ampUrlPrefix` used elsewhere and shouldn't be overloaded? 

Also, does a mapping of `amp-cache-transform` headers to URLs exist somewhere already? Took a look in the amp-packager repo but didn't see anything of note.

sebastianbenz:
`ampUrlPrefix` is meant for runtime self-hosting. In that case you want to rewrite all script srcs, but not all asset links. 

I'm not aware of an existing `amp-cache-transform` header mapping - afaik, the Google AMP Cache is the only one supporting SXG.

twifkak:
[...]

PR #884, opened 2020-08-04
Asset URL rewriting in AMP Optimizer for SXG mode

jonluca:
This is the initial work on implementation of https://github.com/ampproject/amp-toolbox/issues/862

I thought I'd put up the diff in progress to clarify some parts of the problem. 

Based on the past conversation:

> ampUrlPrefix is meant for runtime self-hosting. In that case you want to rewrite all script srcs, but not all asset links.

It looks like `packages/optimizer/lib/transformers/RewriteAmpUrls.js` already rewrites scripts and stylesheets, which should be what's required for SXG compatibility.

Will update this diff as this becomes clearer.


sebastianbenz:
> It looks like packages/optimizer/lib/transformers/RewriteAmpUrls.js already rewrites scripts and stylesheets, which should be what's required for SXG compatibility.

That's correct. One way to handle this could be to init `ampUrlPrefix` based on the sxg option  in `DomTransformer#setConfig`.

The [OptimizeImages transformer](https://github.com/ampproject/amp-toolbox/blob/main/packages/optimizer/lib/transformers/OptimizeImages.js#L163) would be a good place to add the img URL rewriting.

jonluca:
Sounds great! Is there a list of requirements for what other transforms need to be completed? Is rewriting the images to the AMP CDN the last transform? 

Also, just to confirm, the AMP image URLs should be rewritten as such:

`https://i.pinimg.com/236x/b0/7f/1b/b07f1be9f05baad6ec64af0ebeb1cc34.jpg` becomes

`https://i-pinimg-com.cdn.ampproject.org/i/s/i.pinimg.com/236x/b0/7f/1b/b07f1be9f05baad6ec64af0ebeb1cc34.jpg?nii=t`?

As in, we dash-concatenate the domain and subdomain and turn it into a subdomain of the AMP cache, then place the full url as the path in `/i/s/${fullUrl}`?

sebastianbenz:
You can use [toolbox-cache-url](https://github.com/ampproject/amp-toolbox/tree/main/packages/cache-url) to calculate the cache origin. The postfix depends on the protocol. Secure connections have the `/s` appended.

* `https => /i/s/` 
* `http  => /i/`

A few other things which might be helpful:

* You can more or less copy the [implementation from the amppackager](https://github.com/ampproject/amppackager/blob/releases/transformer/transformers/urlrewrite.go#L78) in regards to which attributes need to be rewritten.
* [parseSrcSet](packages/optimizer/lib/parseSrcSet) might be useful for rewriting the srcset attribute.
* we use [cssNano](https://github.com/ampproject/amp-toolbox/blob/main/packages/optimizer/lib/transformers/SeparateKeyframes.js#L22) for parsing CSS.

packages/cli/lib/cmds/optimize.js
```diff
@@ -26,10 +26,12 @@ class OptimizeCmd {
   }
   async run(args, logger) {
     const urlOrPath = args._[1];
-    const {host, rtv, lts, geoapi} = args;
+    const {host, rtv, lts, geoapi, sxgHost, baseDomain} = args;
     const html = await this.load_(urlOrPath);
     const optimized = await this.optimizer_.transformHtml(html, {
       ampUrlPrefix: host,
+      baseDomain: baseDomain,
+      rewriteUrlsToAmpCache: sxgHost,
```
jonluca, line 34: As in the readme, basedomain is the original domain you'd host your content on (for when you pass a relative path to `amp-img`). 

`rewriteUrlsToAmpCache` is going to be the `sxgHost`, i.e. `cdn.ampproject.org`

jonluca:
Sounds good, thanks for the direction! Really helpful! I've updated the diff to address those, and added some basic tests. I think this is getting pretty close to what we need it to do? 

Should we include the CSS parsing here as well?

nainar:
@patrickkettner could you take a look here?

jonluca:
@patrickkettner let me know what you think and if anything should be added!

packages/cli/lib/cmds/optimize.js
patrickkettner: The naming is a bit ambiguous - it reads as an action, rather than a label. would `ampCacheURL` fit a bit better, or am I misunderstanding its purpose?

packages/optimizer/README.md
```diff
@@ -109,6 +111,44 @@ Automatically import any missing AMP Extensions (e.g. amp-carousel).
...
+
+```html
+<img src="/static/img.jpg"/>
+```
+would turn it into
+```html
+<img src="https://example.com/static/img.jpg"/>
+```
+
+#### `fetch`
```
patrickkettner, line 129: thanks for documenting this!

packages/optimizer/lib/transformers/OptimizeAssets.js
```diff
@@ -151,28 +152,84 @@ class OptimizeImages {
...
 
+  async _replaceUrl(url, host) {
+    return createCacheUrl(host, url);
+  }
+
+  async rewriteHost(src, host, baseDomain) {
+    if (!src) {
+      return;
+    }
+    if (src.startsWith('/')) {
```
patrickkettner, line 189: do you want to support relative urls for parent/sibling directories? e.g. `../../foo.jpg`

packages/optimizer/lib/transformers/OptimizeAssets.js
```diff
@@ -151,28 +152,84 @@ class OptimizeImages {
...
+        host,
+        baseDomain
+      );
+    } else if (['use'].includes(tag)) {
+      node.attribs['xlink:href'] = await this.rewriteHost(
+        node.attribs['xlink:href'],
+        host,
+        baseDomain
+      );
+    }
```
patrickkettner, line 230: great work! while you are here, any chance you want to take a crack at background-image, or other css assets?

patrickkettner:
@jonluca would you be able to sign the updated CLA?

PR #891, opened 2020-08-14
Small README fixed

samouri:
- `fetch` was a missing option
- param --> opts in an example.
- outdated/incorrect first option fixed


samouri:
Oh darn. looks like this was merged against master instead of `main`.
It may be worth deleting master so that this can't happen again?

I'll open this PR again against main in the meantime.

PR #892, opened 2020-08-14
Small README fixes

samouri:
Redo of https://github.com/ampproject/amp-toolbox/pull/891, but against main instead of master.


samouri:
While reviewing this, I noticed that `fetch` option is missing its whole description in main

sebastianbenz:
thanks again! I left master to not break any existing links to source files.

PR #899, opened 2020-08-18
Make AMP Toolbox build on windows

sebastianbenz:
This has a few positive sideeffects:

* Normalize eslint config
* Remove linter specific lint task
* This also replaces travis with github actions for testing windows
support in our CI.


package.json
```diff
@@ -14,8 +14,8 @@
     "test:node": "jest",
     "test:optimizer:snapshot": "OPTIMIZER_SNAPSHOT=true jest",
     "test:browser": "karma start --single-run --browsers ChromeHeadless karma.conf.js",
-    "lint": "eslint  -c .eslintrc.json . && lerna run lint",
-    "lint:fix": "eslint  -c .eslintrc.json --fix . && lerna run lint:fix"
+    "lint": "eslint  -c .eslintrc.json .",
+    "lint:fix": "eslint  -c .eslintrc.json --fix '**/*.{js,ts}'"
```
tharders, line 18: Is it ensured that the fix is only run in the source folders and not node_modules and dist folders?

sebastianbenz: We now run eslint from the root dir for all projects. Before linter had it's own eslint config.

packages/linter/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
```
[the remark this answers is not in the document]
sebastianbenz: fixed

jaygray0919:
We are working on a WSL build for amp.dev. Our process is to start with mac/homebrew and then reverse engineer the workflow into WSL. (aside: the first issue we hit was that a gulp install on WSL is different than on homebrew). Are there any lessons-learned or shared methods/libs from this effort that we can leverage to implement amp.dev on WSL?

sebastianbenz:
Unfortunately not. amp.dev is blocked by grow.io which does not work on
windows.

>

PR #901, opened 2020-08-18
Feature/linter optimizer checks

tharders:
Implementing linter page experience checks for
* Page is transformed AMP
* Page is using Java Script Module version of the AMP Runtime.


tharders:
I wonder if we should somehow add the recommendation to the linter rule (for example in the info field).
Otherwise it might not be clear what to do when running the tests in the command line.
(It looks like the "info" is currently not printed for results in the command line)

packages/linter/src/rules/IsTransformedAmp.ts
```diff
@@ -0,0 +1,17 @@
+import { Context } from "../index";
+import { Rule } from "../rule";
+
+export class IsTransformedAmp extends Rule {
+  run({ $ }: Context) {
+    const isTransformed = $("html[transformed='self;v=1']").length > 0;
```
sebastianbenz, line 6: Checking for the presence of `transformed` is enough. The content might change.

packages/linter/src/rules/ModuleRuntimeUsed.ts
```diff
@@ -0,0 +1,27 @@
+import { Context } from "../index";
+import { Rule } from "../rule";
+
+export class ModuleRuntimeUsed extends Rule {
+  run({ $ }: Context) {
+    const isTransformed = $("html[transformed='self;v=1']").length > 0;
```
sebastianbenz, line 6: Might make sense to extract this into a helper method. We'll need it more often.

packages/linter/src/rules/ModuleRuntimeUsed.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (!isTransformed) {
+      return this.pass();
+    }
+    const isModuleVersion =
+      $("script[type='module'][src='https://cdn.ampproject.org/v0.mjs']")
```
sebastianbenz, line 11: Just check for `/v0.mjs` to support runtime self-hosting in the future.

packages/linter/src/rules/ModuleRuntimeUsed.ts
```diff
@@ -0,0 +1,27 @@
...
+        .length > 0;
+    return isModuleVersion
+      ? this.pass()
+      : this.warn(
+          "The Java Script module version of the AMP Runtime is not used"
```
sebastianbenz, line 16: s/Java Script/JavaScript

packages/linter/src/rules/IsTransformedAmp.ts
tharders: done

packages/linter/src/rules/ModuleRuntimeUsed.ts
tharders: done

packages/linter/src/rules/ModuleRuntimeUsed.ts
tharders: done

packages/linter/src/rules/ModuleRuntimeUsed.ts
tharders: done

sebastianbenz:
Build failure is unrelated

PR #902, opened 2020-08-19
Preload checks for blocking extensions

tharders:
Preload check for runtime now also allows hosted version and javascript module version.
Do we need to change the warning message now?


packages/linter/src/rules/BlockingExtensionsPreloaded.ts
```diff
@@ -0,0 +1,27 @@
+import { Context } from "../index";
+import { Rule } from "../rule";
+
+const blockingExtension = ["amp-experiment", "amp-dynamic-css-classes"];
```
sebastianbenz, line 4: amp-story needs to go here as well.

packages/linter/src/rules/RuntimeIsPreloaded.ts
```diff
@@ -3,14 +3,13 @@ import { Rule } from "../rule";
...
-    const attr = [
-      "href='https://cdn.ampproject.org/v0.js'",
-      "rel='preload'",
-      "as='script'",
-    ]
+    const jsAttr = ["href$='/v0.js'", "rel='preload'", "as='script'"]
       .map((s) => `[${s}]`)
       .join("");
-    const isPreloaded = $(`link${attr}`).length > 0;
+    const mjsAttr = ["href$='/v0.mjs'", "rel='modulepreload'"]
```
sebastianbenz, line 9: We're currently not using `modulepreload` so please omit checking for this attribute

packages/linter/src/rules/BlockingExtensionsPreloaded.ts
tharders: done

packages/linter/src/rules/RuntimeIsPreloaded.ts
tharders: For mjs we now simply check if the rel attribute ends with preload, so normal preload (as it is used now) or modulepreload will be fine.

sebastianbenz (approved):

sebastianbenz:
Thanks!

PR #909, opened 2020-08-25
Fix handling of malformed `sizes` and `heights` strings

chasefinch:
Some of the test output files for server-side rendering contained malformed css.

The comment above the test element said "removes invalid sizes", but in `parseSizesSpec.js` it seems that invalid sizes definition should fail in general. I can't think of reasonable logic to salvage the invalid sizes in the test input files, so I added the invalid sizes string as a test in parseSizesSpec.js, and removed it from the output files.


packages/optimizer/spec/transformers/valid/ServerSideRendering/converts_sizes_attribute_to_css/input.html
```diff
@@ -25,14 +25,5 @@
       src=https://acme.org/image1.png
       width=400>
     </amp-img>
-    <!-- removes invalid sizes -->
```
sebastianbenz, line 28: Is this now covered by a different test case?

packages/optimizer/spec/transformers/valid/ServerSideRendering/converts_heights_attribute_to_css/input.html
```diff
@@ -15,13 +15,5 @@
       src=https://acme.org/image1.png
       width=320>
     </amp-img>
-    <!-- removes invalid heights -->
```
sebastianbenz, line 18: Is this case covered by a different test case?

sebastianbenz (commented):
Thanks! I think it's still important to test the invalid case in the transformer tests, just to be sure that this won't break the transformer.

packages/optimizer/spec/transformers/valid/ServerSideRendering/converts_sizes_attribute_to_css/input.html
sebastianbenz: Ignore previous comment, however, I'd prefer to keep this.

packages/optimizer/spec/transformers/valid/ServerSideRendering/converts_heights_attribute_to_css/input.html
sebastianbenz: ignore comment

chasefinch:
Leaving the invalid tests in causes the expected output to preserve the boilerplate from the input; Is that okay?

sebastianbenz:
Good point. Then we should move it to a different file.

sebastianbenz (approved):
Thanks a lot!

PR #911, opened 2020-08-27
Allow users to specify version numbers for auto imported extensions

digijin:
AutoExtensionImport adds the latest versions of extensions by default. Some projects/teams want more control so they can pin to a specific tested version and not worry about unexpected changes as new versions come out, or so they can manually run their tests before upgrading. Similar to the problem that yarn and npm use lockfiles to fix.

This PR allows a new config parameter to be passed in which describes the version number to be used for each extension.

### DIFF
--- a/packages/optimizer/lib/transformers/AutoExtensionImporter.js
+++ b/packages/optimizer/lib/transformers/AutoExtensionImporter.js
@@ -60,6 +60,7 @@ class AutoExtensionImporter {
     this.format = config.format || DEFAULT_FORMAT;
     this.log_ = config.log.tag('AutoExtensionImporter');
     this.experimentBindAttributeEnabled = config.experimentBindAttribute === true;
+    this.extensionVersions = config.extensionVersions || {};
   }
 
   /**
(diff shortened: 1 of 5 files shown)


packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -60,6 +60,7 @@ class AutoExtensionImporter {
     this.format = config.format || DEFAULT_FORMAT;
     this.log_ = config.log.tag('AutoExtensionImporter');
     this.experimentBindAttributeEnabled = config.experimentBindAttribute === true;
+    this.extensionVersions = config.extensionVersions || {};
```
sebastianbenz, line 63: Theoretically, we could validate the provided parameter by: 

1. matching the extensions against the ones on the validator rules
2. checking if the version number is supported by the extension

sebastianbenz (approved):
Very nice! Thanks a lot! Left one suggestion.

packages/optimizer/lib/transformers/AutoExtensionImporter.js
james-smith-foxsports: I would argue this shouldn't be validated, at least not in terms of ensuring the version number is supported. If I was using an invalid version number then either I'm doing it for a specific reason, or its clearly my fault and I should be able to debug from seeing the script 404 in my browser. In the case where I am doing it for a specific reason (e.g. playing with self hosting amp scripts) then I wouldn't want to deal with validation errors, I would expect the code to be "dumb" and just do what I tell it.

sebastianbenz: sgtm

PR #913, opened 2020-08-28
Use node instead of root for checking templates in PreloadHeroImages

schlessera:
In the `PreloadHeroImages` transformer's `findHeroImages()` method, the template check is always executed against the root `<body>` node, instead of the node currently being processed.

This PR swicthes the check from `root` to `node`.

### DIFF
--- a/packages/optimizer/lib/transformers/PreloadHeroImage.js
+++ b/packages/optimizer/lib/transformers/PreloadHeroImage.js
@@ -137,7 +137,7 @@ class PreloadHeroImage {
       if (!heroImageCandidate && heroImages.length === 0) {
         heroImageCandidate = this.isCandidateHeroImage(node);
       }
-      if (amphtml.isTemplate(root)) {
+      if (amphtml.isTemplate(node)) {
         // Ignore images inside templates
         node = skipNodeAndChildren(node);
       } else {


schlessera:
Note: the pre-push checks failed for seemingly unrelated spec test failures. I skipped them to be able to create this PR, but I'm not sure what that entails.

patrickkettner:
hi @schlessera!

forgive me, but [aren't they the same](https://github.com/ampproject/amp-toolbox/blob/b3ed0b9c28c7af7a08a72a493a7b2c19fb8b75a9/packages/optimizer/lib/transformers/PreloadHeroImage.js#L131)?

schlessera:
@patrickkettner No, they are only the same for the first iteration of the following loop.

The loop basically does this:
```js
while (node !== null) {
   if (amphtml.isTemplate(node)) {
      node = skipNodeAndChildren(node);
   } else {
      node = nextNode(node);
   }
}
```

So it iterates over nodes while sometimes skipping children.

However, while doing so, the logic within the loop always checks `root` instead of `node`, which stays the same over all iterations of the loop.

So it does the exact same check on the initial root node multiple times and fails to check any of the subsequent nodes that are being looped over.

Note: I might be missing something from the fact I'm not a specialized JS developer. But the way I understand the code, both `root` and `node` will initially be copies of the same reference pointing to the same node. However, during the loop, we're not using that reference to modify the object (which would be the same for both), we're actually swapping the reference for a new one. So `node` will be updated to point to a new object, but `root` will not.

patrickkettner:
ah! of course. thanks!

PR #914, opened 2020-08-29
Remove unneeded width & height checks

schlessera:
The `isTinyNode()` method first ensures that all negative or zero values of either `width` or `height` are eliminated.

Then, after a check for responsiveness, it checks the `width` and `height` again for being a positive value. At his point, this is guaranteed to be true already by the first check, and thus only adds unneeded processing overhead.

This PR removes these unneeded checks, assuming that the initial check is correct.


patrickkettner:
thanks!

PR #918, opened 2020-09-08
Add missing closing tags to <amp-img> elements

### LINKED ISSUE #917: Spec file missing closing <amp-img> tags
Here's one example where the `PreloadHeroImage` transformer messes up the DOM and inadvertently moves the second `<amp-img>` element to be a child of the first instead of its sibling:

https://github.com/ampproject/amp-toolbox/blob/d0f4621464218a3a3a5a1f50cf3436521aef1b45/packages/optimizer/spec/transformers/valid/PreloadHeroImage/srcset_cannot_be_preloaded/expected_output.html#L5-L9

~My initial suspicion is that this is due to combining DOM iteration with DOM manipulation in one go, which will cause the DOM manipulations (adding an element) to shift what the DOM iterator is pointing at (but that's just a guess for now).~

The problem is actually that the `input.html` file is already corrupt and is missing the closing tags for the `<amp-img>` elements - PR incoming.

schlessera:
The `input.html` file for the `PreloadHeroImage/srcset_cannot_be_preloaded` spec test is missing the closing tags for its `<amp-img>` elements. This causes the DOM to get corrupted when imported and the `<amp-img>` elements to be inadvertently nested.

Fixes #917

### DIFF
--- a/packages/optimizer/spec/transformers/valid/PreloadHeroImage/srcset_cannot_be_preloaded/expected_output.html
+++ b/packages/optimizer/spec/transformers/valid/PreloadHeroImage/srcset_cannot_be_preloaded/expected_output.html
@@ -2,9 +2,7 @@
 <head></head>
 <body>
   <!-- srcset is not supported by all browsers -->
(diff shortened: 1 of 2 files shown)


patrickkettner:
thanks!

PR #929, opened 2020-09-25
Remove unneeded extra semicolon

schlessera:
There's a trailing semi-colon being added to the inline style of the `i-amphtml-sizer`. This just adds wasted bytes to the HTML and should be removed.


schlessera:
Closed in favor of #930 (to avoid merge conflicts)

PR #930, opened 2020-09-25
Avoid trailing zeros and semicolons in sizer's padding

schlessera:
The inline style for responsive sizers currently wastes bytes by adding fixed trailing zeroes to its value, like `80.0000%`. This PR fixes this to only use as many digits as needed for values like this, stripping any unneeded zeroes and maybe even the period.

It also strips the trailing semi-colon that is not needed after the padding to save one more byte per sizer.

### DIFF
--- a/packages/optimizer/lib/transformers/ApplyLayout.js
+++ b/packages/optimizer/lib/transformers/ApplyLayout.js
@@ -118,7 +118,7 @@ function maybeAddSizerInto(node, layout, width, height) {
 function createResponsiveSizer(width, height) {
   const padding = (height.numeral / width.numeral) * 100;
   const sizer = createElement('i-amphtml-sizer', {
-    style: `display:block;padding-top:${padding.toFixed(4).replace(/\.?0*$/, '')}%`,
+    style: `display:block;padding-top:${parseFloat(padding.toFixed(4))}%`,
   });
   return sizer;
 }
--- a/packages/optimizer/lib/transformers/ApplyLayout.js
+++ b/packages/optimizer/lib/transformers/ApplyLayout.js
@@ -118,7 +118,7 @@ function maybeAddSizerInto(node, layout, width, height) {
 function createResponsiveSizer(width, height) {
   const padding = (height.numeral / width.numeral) * 100;
   const sizer = createElement('i-amphtml-sizer', {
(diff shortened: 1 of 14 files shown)


packages/optimizer/lib/transformers/ApplyLayout.js
```diff
@@ -118,7 +118,7 @@ function maybeAddSizerInto(node, layout, width, height) {
 function createResponsiveSizer(width, height) {
   const padding = (height.numeral / width.numeral) * 100;
   const sizer = createElement('i-amphtml-sizer', {
-    style: `display:block;padding-top:${padding.toFixed(4)}%;`,
+    style: `display:block;padding-top:${padding.toFixed(4).replace(/\.?0*$/, '')}%`,
```
patrickkettner, line 121: any reason we aren't using `parseFloat`?

schlessera: Well, we already have a float, the problem is to format a float correctly into a string, not the other way around.

patrickkettner: well anything used here has toString implicitly being called on it  as a result of being referenced inside of a template literal.

Taking a float, `toFixed`-ing it, and then parsing it again would accomplish the same thing without a regex, right?

schlessera: That seems to be slower than using a regex, though:

Image 2020-10-07 at 8 14 16 AM

See https://www.measurethat.net/Benchmarks/Show/9870/0/float-string-optimization-parsefloat-vs-regex

patrickkettner: for sure! and great research, but it was more about the legibility of the code rather than the performance of several million parses. I am hard-pressed to think of a situation where that would be plausible

schlessera: Got it, I'll adapt the code.

schlessera: Resolved via https://github.com/ampproject/amp-toolbox/pull/930/commits/6ad6c9eac78caf49d57e23964049ca834ee92f0e

patrickkettner:
@schlessera terribly sorry, would you be able to sign off on the new CLA? thanks!

PR #936, opened 2020-10-12
Batch wrapper updates

Dbrown910:
Added a new image size calculator package. Done to prevent timeouts when running the linter programmatically in a wrapper. Previous image-size library was built on HTTP and subsequent requests to the same domain for assets would error out


tharders (changes requested):
Hi Damani, thank you for your fix.
The drawback here is that now the complete image is downloaded into a buffer, which may be quite big while the probe-image-size library only streams the response until the image headers are loaded.
It is not clear to me what the original issue was and how to reproduce it.
Do you have a test case that reproduces the problem with probe-image-size?
If so we might have a good chance the issue is fixed on the probe-image-size side.
We could also try to play around with some request options in order to fix the problem.

PR #952, opened 2020-11-19
Fix various issues with blurry image placeholder transform

### LINKED ISSUE #950: Problem with enable Blurry image placeholders
We are trying to implement Blurry image placeholders, but we cannot see any changes. Our integration is developed through Lambda Egde.

The current configuration is:
`
const ampOptimizer = AmpOptimizer.create({
        preloadHeroImage: false,
        transformations: AmpOptimizer.TRANSFORMATIONS_PAIRED_AMP,
        blurredPlaceholders: true
});
`
It is running on the following [link](https://vader.news/.tests/ssr/noticia_0005.html)

Question: Currently not supported for valid AMP pages? We try to have it inline too, but it throws amp validation error. Any roadmap that is supported for valid amp?

Thanks!

jridgewell:
- Add `AddBlurryImagePlaceholders` to AMP_FIRST transformers, since it's now valid in transformed output
- Adds `i-amphtml-ssr` attribute to the `<amp-img>` node when adding a blurry image (it's required for the transformed output to be valid)
- Adds a basic `User-Agent` header to requests so that we're compatible with [Cloudflare WAF](https://www.cloudflare.com/lp/waf-x/)

Fixes #950


sebastianbenz (approved):
Nice! Thanks! Now we only need to switch from jimp to sharp.

PR #961, opened 2020-11-30
Upgrade dependencies 

sebastianbenz:


packages/linter/src/helper.ts
```diff
@@ -73,7 +74,7 @@ export function fetchToCurl(
   const headers = init.headers || {};
 
   const h = Object.keys(headers)
-    .map((k) => `-H '${k}: ${headers[k]}'`)
```
patrickkettner, line 76: it looks like you removed a few `()` around single arguments. Would it make sense to add `arrowParens: "avoid"` to  the prettier [config](https://github.com/ampproject/amp-toolbox/blob/main/prettier.config.js)?

packages/linter/src/index.ts
```diff
@@ -30,33 +30,34 @@ import { HeroImageIsDefined } from "./rules/HeroImageIsDefined";
 import { FastGoogleFontsDisplay } from "./rules/FastGoogleFontsDisplay";
 import { GoogleFontPreconnect } from "./rules/GoogleFontPreconnect";
 import { BoilerplateIsRemoved } from "./rules/BoilerplateIsRemoved";
-import { AmpImgUsesSrcSet } from './rules/AmpImgUsesSrcSet';
```
patrickkettner, line 33: same prettier comment as above (we actually have [singleQuotes](https://prettier.io/docs/en/options.html#quotes) as `true` [currently](https://github.com/ampproject/amp-toolbox/blob/cff4fc0f1c5548e1b39d15d737ad530ee22459dd/prettier.config.js#L19))

patrickkettner: er, actually running prettier on the file updated the quotes - are we just not including prettier in the precommit hook?

packages/linter/src/index.ts
```diff
@@ -30,33 +30,34 @@ import { HeroImageIsDefined } from "./rules/HeroImageIsDefined";
...
 import { isArray } from "util";
+import * as cheerio from "cheerio";
 
 export enum LintMode {
   Amp = "amp",
   AmpStory = "ampstory",
   Amp4Ads = "amp4ads",
   Amp4Email = "amp4email",
   PageExperience = "pageexperience",
-  Sxg = "sxg",
```
patrickkettner, line 43: should [`es5`](https://github.com/ampproject/amp-toolbox/blob/cff4fc0f1c5548e1b39d15d737ad530ee22459dd/prettier.config.js#L21) be changed to [`none`](https://prettier.io/docs/en/options.html#trailing-commas)?

packages/linter/tests/local/TitleMeetsLengthCriteria-2/source.html
```diff
@@ -16,7 +16,7 @@
...
    </head>
    <body>
-      <amp-story standalone poster-portrait-src="https://ss.makestories.io/get?story=-M7d0U5_iOxv_iRRUsOg&version=1589826244967&width=696&height=1237" publisher-logo-src="https://app.makestories.io/images/favicon.png" publisher="MakeStories" title="Terrible Title: The Story Of How I Rambled To Make This Forty Characters Long">
+      <amp-story standalone poster-portrait-src="https://ss.makestories.io/get?story=-M7d0U5_iOxv_iRRUsOg&version=1589826244967&width=696&height=1237" publisher-logo-src="https://app.makestories.io/images/favicon.png" publisher="MakeStories" title="Terrible Title: The Story Of How I Rambled To Make This Ninty Characters Long...................................">
```
patrickkettner, line 19: ever the wordsmith :D

packages/optimizer-docker/utils.js
```diff
@@ -1,5 +1,9 @@
 const url = require('url');
 
+function objectFromEntries(arr) {
```
patrickkettner, line 3: why was Object.fromEntries not usable? Would make sense to explain in a comment, because I feel like I would switch this back without seeing it was removed first

patrickkettner (approved):
LGTM
as noted, a lot of the formatting changes should be automated through `prettier`, but that can be a separate CL

packages/linter/src/index.ts
sebastianbenz: looks like it, we need to fix that

PR #964, opened 2020-12-01
run prettier in precommit hook

patrickkettner:


sebastianbenz (approved):
Thanks! Originally, I thought that eslint would apply the prettier rules during fix.

PR #976, opened 2020-12-07
Check if "." before extension in image URL

### LINKED ISSUE #977: isPathNameAnImage incorrectly identifies image paths
URLs could end in with a string from `imageExtensions ` but not be an image file.
Discovered from [#21473](https://github.com/ampproject/amphtml/issues/31473).

https://github.com/ampproject/amp-toolbox/blob/0c8755016ae825b11b63b98be83271fd14cc0486/packages/cache-url/lib/ImageExtensions.js#L146-L152

processprocess:
Context / Fixes #977

URLs can end with a string in `imageExtensions ` and not be an image. For example:
https://arts.visualstories.com/quizzes/know-your-art-a-trivia-on-renaissance-art 

This will check it the url also has an `.` before the extension. 

For a follow up, we should to do something like this for `isPathNameAFont` too.

### DIFF
--- a/packages/cache-url/lib/ImageExtensions.js
+++ b/packages/cache-url/lib/ImageExtensions.js
@@ -145,7 +145,7 @@ const imageExtensions = [
 const exports = {
   isPathNameAnImage: (filePath) => {
     return imageExtensions.some((extension) => {
-      if (filePath.endsWith(extension)) {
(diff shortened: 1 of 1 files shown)


patrickkettner:
thanks!

PR #979, opened 2020-12-09
Check if "." before extension in font URL

### LINKED ISSUE #978: isPathNameAFont incorrectly identifies image paths
URLs could end in with a string from `fontExtensions` but not be a font file.
Such as a url ending in `museum-collection` would be misidentified as a font file in the current implementation.

https://github.com/ampproject/amp-toolbox/blob/main/packages/cache-url/lib/FontExtensions.js#L220-L227

Discovered from [#21473](https://github.com/ampproject/amphtml/issues/31473).
Follow up to #977

processprocess:
Context / Fixes #978

### DIFF
--- a/packages/cache-url/lib/FontExtensions.js
+++ b/packages/cache-url/lib/FontExtensions.js
@@ -219,7 +219,7 @@ const fontExtensions = [
 const exports = {
   isPathNameAFont: (filePath) => {
     return fontExtensions.some((extension) => {
-      if (filePath.endsWith(extension)) {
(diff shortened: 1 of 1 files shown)


processprocess:
cc @Enriqe 
This should fully remove the bug discovered from [31473](https://github.com/ampproject/amphtml/issues/31473)

patrickkettner:
thanks!

Enriqe:
thanks for taking this @processprocess !

PR #980, opened 2020-12-10
Add check for disabled tap delay

sebastianbenz:
See https://buganizer.corp.google.com/issues/174835448#comment11 and https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away


sebastianbenz:
Not sure though whether we need to check for CSS as well.

patrickkettner (approved):

westonruter:
Looks like this won't be needed anymore soon as the issue was just fixed in Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1108987#c15

PR #981, opened 2020-12-14
Encode SVG sizer with base64

### LINKED ISSUE #844: Use base64 encoding for sizer SVGs
As a follow-up to https://github.com/ampproject/amphtml/pull/27518, the sizer SVGs should default to using base64 encoding to be more robust against HTML minification.

Maybe this could be made configurable, so that you can opt to have the raw HTML representation instead, but I think the base64 encoding is the safer default.

schlessera:
Fixes #844


sebastianbenz (approved):
Great - thanks a lot!

PR #1065, opened 2020-12-15
Split preconnect and dns-prefetch into two separate links

### LINKED ISSUE #960: Preconnect/dns-prefetch resource hint against web.dev best practices
I think this combined preconnect and dns-prefetch resource hint is not always doing what it is supposed to do depending on the browser:

https://github.com/ampproject/amp-toolbox/blob/84cb3057e5f6c54d64369ddd285db1cb36237ee8/packages/optimizer/lib/transformers/GoogleFontsPreconnect.js#L55

As far as I can tell, the `dns-prefetch` is meant to be the fallback in case the `preconnect` is not supported (as the `preconnect` would include a DNS resolution). However, the `dns-prefetch` is mentioned first, which seems odd.

Also, according to web.dev, best practice is to use two separate hints with the `dns-prefetch` as the second hint serving as a fallback, as using a combined hint will break in Safari:
https://web.dev/preconnect-and-dns-prefetch/#resolve-domain-name-early-with-reldns-prefetch

schlessera:
Fixes #960

### DIFF
--- a/packages/optimizer/lib/transformers/GoogleFontsPreconnect.js
+++ b/packages/optimizer/lib/transformers/GoogleFontsPreconnect.js
@@ -65,7 +65,7 @@ class GoogleFontsPreconnect {
         this.log_.debug(
           'adding <link rel="preconnect" href="' + linkPreconnect.attribs.href + '" crossorigin>'
         );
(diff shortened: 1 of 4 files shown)


sebastianbenz (approved):
Thanks!

PR #1081, opened 2020-12-15
Adapt markdown spec output to new base64 encoding of SVGs

schlessera:
This fixes a couple of spec test `expected_output.html` files that were missed in PR #981.


sebastianbenz:
Fixed on main

Issue #959, opened 2020-11-26
AMP Optimizer: Elements with the heights or sizes attributes produces invalid AMP when no layout is present.


Artmann:
When server-side rendering an document, the optimizer will have to determine the layout to use if none is present. Having a layout is required to be able to [generate the correct styles for the element](https://github.com/ampproject/amp-toolbox/blob/556374697b1e1882dccaa8fa00c5a7c9ed265c74/packages/optimizer/lib/transformers/ApplyLayout.js#L61
).

If the element has a `width`, `height` attribute specified and either a `sizes` or `heights` attribute, the `responsive` layout will be [selected](https://github.com/ampproject/amp-toolbox/blob/556374697b1e1882dccaa8fa00c5a7c9ed265c74/packages/optimizer/lib/ParseLayout.js#L129).

This will add the classes `i-amphtml-layout-responsive` and `i-amp-layout-size-defined` as well as the attribute `i-amp-layout=responsive`. This is does not pass validation when the `layout=responsive` attribute is missing.

For Example, this input

```
<amp-ad
  class=""
  type=taboola
  data-block-on-consent=_till_responded
  width=300
  height=250
  heights="(min-width:616px) 370%, (min-width:472px) 386%, (min-width:375px) 411%, 438%"
  data-publisher=...
  data-mode=thumbnails-a-amp
  data-placement="Mobile Below Article Thumbnails AMP"
  data-article=auto
  data-target_type=mix
>
  ...
</amp-ad>
```

becomes

```
<amp-ad
  class=" i-amphtml-layout-responsive i-amphtml-layout-size-defined"
  type=taboola
  data-block-on-consent=_till_responded
  width=300
  height=250
  data-publisher=diariodebiologia
  data-mode=thumbnails-a-amp
  data-placement="Mobile Below Article Thumbnails AMP"
  data-article=auto
  data-target_type=mix
  i-amphtml-layout=responsive
  id=i-amp-0
>
  <i-amphtml-sizer style="display:block;padding-top:83.3333%;"></i-amphtml-sizer>
</amp-ad>
```

Which doesn't pass AMP Validation. ❌

christofferartmannmrf:
Thanks for the help. ❤️

sebastianbenz:
No problem. Thanks for the detailed bug report!

PR #1115, opened 2020-12-16
Auto add layout=responsive if missing

sebastianbenz:
Fixes https://github.com/ampproject/amp-toolbox/issues/959

### DIFF
--- a/packages/optimizer/lib/transformers/ApplyLayout.js
+++ b/packages/optimizer/lib/transformers/ApplyLayout.js
@@ -184,7 +184,6 @@ module.exports = {
     // Transformed AMP validation requires layout attribute to be set
     // See https://github.com/ampproject/amp-toolbox/issues/959
     if (layout && layout === 'responsive') {
-      console.log('XXX setting custom layout ', layout);
       customElement.attribs.layout = layout;
     }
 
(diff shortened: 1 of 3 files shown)


patrickkettner (approved):

packages/optimizer/lib/transformers/ApplyLayout.js
```diff
@@ -181,6 +181,12 @@ module.exports = {
       log.debug('cannot perform SSR: unsupported layout', layout);
       return false;
     }
+    // Transformed AMP validation requires layout attribute to be set
+    // See https://github.com/ampproject/amp-toolbox/issues/959
+    if (layout && layout === 'responsive') {
+      console.log('XXX setting custom layout ', layout);
```
patrickkettner, line 187: why `console.log` vs the `log` util?

patrickkettner:
LGTM 

wondering why responsive is the assumed default, and if there is a chance that having a different configurable default would make sense?

packages/optimizer/lib/transformers/ApplyLayout.js
sebastianbenz: removed

sebastianbenz:
`responsive` is not the default. It's only that the validator does not auto infer the responsive layout as the server-side-rendering does.

PR #1117, opened 2020-12-22
add cross-fetch back

patrickkettner:
it is breaking older versions that have a transient dependency on cross-fetch, namely nextjs


sebastianbenz:
Fixed by #1118

PR #1118, opened 2020-12-23
Readd cross fetch

sebastianbenz:
This broke older versions of next.js.


patrickkettner (approved):
LGTM assuming it passes

sebastianbenz:
That's the problem...works locally, but doesn't pass in CI. Not sure what the problem is.

patrickkettner:
Yeah, I wasn't been able to figure it out all day yesterday￼.

PR #1124, opened 2021-01-09
Only add space between classes if preexisting class is present

### LINKED ISSUE #1123: Superfluous leading space introduced in element class
When the layout classes are added to an element, a superfluous leading space is added if the class attribute already existed but was empty.

See as an example:
https://github.com/ampproject/amp-toolbox/blob/53951ba01cad5d7724296f23254ed6b341baa86e/packages/optimizer/spec/transformers/valid/ServerSideRendering/transforms_layout_responsive/expected_output.html#L10

schlessera:
A change in https://github.com/ampproject/amp-toolbox/pull/1115 surfaced a problem within `addClass()` helper function, where an existing but empty `class` attribute leads to a superfluous prefixing space.

This PR changes the code in `addClass()` to account for this special case and adapts the expected spec output as required.

Fixes #1123

### DIFF
--- a/packages/optimizer/lib/transformers/ApplyLayout.js
+++ b/packages/optimizer/lib/transformers/ApplyLayout.js
@@ -48,7 +48,8 @@ function getAttributeOrNull(element, name) {
 }
 
 function addClass(node, value) {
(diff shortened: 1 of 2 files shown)


sebastianbenz (approved):
Nice! Thanks!

PR #1125, opened 2021-01-11
Support script based templates

sebastianbenz:
Make sure to always also support the script based template syntax:

```
<script type="text/plain" template="amp-mustache">
  Hello {{world}}!
</script>
```

Using the template helper should also make SSR more performant as it
avoids a full parent traversal for each node.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -20,12 +20,6 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
           registry-url: https://registry.npmjs.org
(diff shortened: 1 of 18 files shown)


patrickkettner (commented):
LGTM - I can't find this documented anywhere, however. Is there any?

sebastianbenz:
It's documented here: https://amp.dev/documentation/components/amp-mustache/.

patrickkettner:
oh, pft, I somehow changed it to amp-script powered templates when I was looking for docs. apologies.

PR #1128, opened 2021-01-14
account for cases when parentNode and nextNode are undefined

patrickkettner:
fix for #1121

### DIFF
--- a/packages/optimizer/package.json
+++ b/packages/optimizer/package.json
@@ -42,7 +42,6 @@
     "htmlparser2": "5.0.1",
     "https-proxy-agent": "5.0.0",
     "lru-cache": "6.0.0",
-    "mock-require": "^3.0.3",
     "normalize-html-whitespace": "1.0.0",
     "postcss": "7.0.32",
     "postcss-safe-parser": "4.0.2",
--- a/packages/optimizer/package.json
+++ b/packages/optimizer/package.json
@@ -42,6 +42,7 @@
     "htmlparser2": "5.0.1",
     "https-proxy-agent": "5.0.0",
     "lru-cache": "6.0.0",
+    "mock-require": "^3.0.3",
     "normalize-html-whitespace": "1.0.0",
     "postcss": "7.0.32",
     "postcss-safe-parser": "4.0.2",
--- a/packages/optimizer/spec/helpers/TransformerRunner.js
+++ b/packages/optimizer/spec/helpers/TransformerRunner.js
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 const validator = require('amphtml-validator');
+const mockRequire = require('mock-require');
 const {basename, join} = require('path');
 const {writeFileContents, getFileContents, getDirectories} = require('../helpers/Utils.js');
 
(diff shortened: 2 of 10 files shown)


packages/optimizer/spec/lib/HtmlDomHelper.test.js
```diff
@@ -28,6 +28,21 @@ test('findMetaViewport returns null if tag is not present', async () => {
   expect(result).toBeNull();
 });
 
+test('works even when template is the final node', async () => {
+  let err = null;
+
+  try {
+    await treeParser.parse(`<!doctype html>
```
sebastianbenz, line 35: Does this test ever fail?

sebastianbenz (commented):
Can we test that non of the transformers fails when skipNodeAndChildren returns null?

packages/optimizer/spec/lib/HtmlDomHelper.test.js
patrickkettner: it would fail if the code I added isn't there

patrickkettner:
> Can we test that non of the transformers fails when skipNodeAndChildren returns null?

will do

patrickkettner:
@sebastianbenz I have gotten stuck on the tests. I am trying to reuse the same looped abstraction to test all of the transformers, and just unregister, mock, and reregister. But I am getting lost in jest - I pushed my current (broken) state, would you be able to take a look?

packages/optimizer/spec/helpers/TransformerRunner.js
```diff
@@ -45,56 +46,81 @@ if (WRITE_SNAPSHOT) {
...
         }
-        done();
+      }
+      done();
+    };
+
+    getDirectories(testConfig.testDir).forEach((testDir) => {
+      it(basename(testDir), (done) => runTests(testDir, done));
+
+      it('works when skipNodeAndChildren returns null', function (done) {
```
sebastianbenz, line 103: I'd prefer not to add more tests here. Instead, we could add an end-to-end test [here](https://github.com/ampproject/amp-toolbox/tree/master/packages/optimizer/spec/end-to-end) that runs a document with a template as last element through all transformers.

packages/optimizer/spec/lib/HtmlDomHelper.test.js
sebastianbenz: ok, interesting. skipNodeAndChildren is only called from within transformers and these shouldn't be run when calling treeParser.parse.

packages/optimizer/spec/helpers/TransformerRunner.js
patrickkettner: sounds good

packages/optimizer/package.json
```diff
@@ -42,6 +42,7 @@
     "htmlparser2": "5.0.1",
     "https-proxy-agent": "5.0.0",
     "lru-cache": "6.0.0",
+    "mock-require": "^3.0.3",
```
sebastianbenz, line 45: Do we still need this?

sebastianbenz (approved):
Thanks!

sebastianbenz:
Can you revert the package lock to main.

sebastianbenz:
Thanks Patrick! The CI failures were unrelated to this PR.

PR #1132, opened 2021-01-29
Don't preload hero images

sebastianbenz:
We can't tell whether a hero image is responsive and is loaded depending
on viewport size. The current implementation will always preload hero
images, which means images might be preloaded even if they are not
visible on the page. This PR reverts this behavior and disables
preload generation for hero images.

The only exception is when an image defines a media attribute.

### DIFF
--- a/packages/optimizer/lib/transformers/PreloadHeroImage.js
+++ b/packages/optimizer/lib/transformers/PreloadHeroImage.js
@@ -101,9 +101,12 @@ class PreloadHeroImage {
       'as': 'image',
       'data-hero': '',
     });
-    if (heroImage.media) {
-      preload.attribs.media = heroImage.media;
+    if (!heroImage.media) {
+      // We can only safely preload a hero image if there's a media attribute
+      // as we can't detect whether it's hidden on certain viewport sizes otherwise.
+      return;
     }
+    preload.attribs.media = heroImage.media;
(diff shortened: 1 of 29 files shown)


packages/optimizer/lib/transformers/PreloadHeroImage.js
```diff
@@ -101,9 +101,12 @@ class PreloadHeroImage {
       'as': 'image',
       'data-hero': '',
     });
-    if (heroImage.media) {
-      preload.attribs.media = heroImage.media;
+    if (!heroImage.media) {
+      // We can only safely preload a hero image if there's a media attribute
+      // as we can't detect whether it's hidden on certain viewport sizes otherwise.
```
patrickkettner, line 106: Why? Is it an opt in sort of thing, or does the media attr provide some kind of indication of..something?

patrickkettner (approved):
lgtm

packages/optimizer/lib/transformers/PreloadHeroImage.js
sebastianbenz: This is AMP's built in way to define per element media queries. The assumption is that if you use this, you won't define additional media queries in CSS. This could still lead to false positives, but I suspect that this is seldom the case.

PR #1135, opened 2021-02-03
Add missing amp-ad ending tag

schlessera:
The `input.html` file in the `ServerSideRendering/transforms_layout_responsive` spec test is missing an ending tag for one of the `<amp-ad>` elements, which messes up the test output.


sebastianbenz (approved):
Thanks for the fix!

PR #1137, opened 2021-02-11
Drop the final width descriptor in a generated srcset.

kristoferbaxter:
Based on discoveries in https://github.com/kristoferbaxter/srcset-broken

Leaving this PR in draft form while investigating further.


patrickkettner (approved):

sebastianbenz:
I'm going to merge this for now so that we can experiment with it on amp.dev.

PR #1140, opened 2021-02-11
Enable ESM module version of AMP Runtime and Components by default

sebastianbenz:


packages/optimizer/README.md
```diff
@@ -111,6 +112,20 @@ Automatically import any missing AMP Extensions (e.g. amp-carousel).
 - default: `true`
 - used by: [AutoExtensionImport](lib/transformers/AutoExtensionImporter.js)
 
+#### `esmModulesEnabled`
+
+Enables the smaller ESM module version of AMP runtime and components:
```
patrickkettner, line 117: This is a bit ambiguous - is there a way to link to some kind of docs to explain this more?

packages/optimizer/lib/transformers/RewriteAmpUrls.js
```diff
@@ -56,7 +56,7 @@ const {calculateHost} = require('../RuntimeHostHelper');
  */
 class RewriteAmpUrls {
   constructor(config) {
-    this.esmModulesEnabled = config.experimentEsm;
```
patrickkettner, line 59: Since you aren't supporting the old version at all, this be a breaking change as far as semver is concerned, right?

patrickkettner (approved):

packages/optimizer/lib/transformers/RewriteAmpUrls.js
sebastianbenz: Good question. My assumption is: the previous default was disabled, the new default is enabled. 

Cases:

* If someone previously enabled it, it's still going to be enabled => not a breaking change
* If someone previously explicitly disabled it, it's going to be enabled now. This is in theory a breaking change, however, this is the new "official" default for AMP as it's valid AMP now. And the flag was named experimental + undocumented, so I wouldn't consider it part of the official API.

packages/optimizer/README.md
sebastianbenz: Not yet - we should add one...once there's one.

PR #1142, opened 2021-02-12
Don't remove sizes for hero images

sebastianbenz:
Workaround for https://github.com/ampproject/amphtml/issues/32644.


sebastianbenz:
Merging to get the fix in due to US holiday.

PR #1143, opened 2021-02-18
Add extension scripts in logical order

schlessera:
When an extension is added by the `AutoExtensionsImporter` in the current code, the `referenceNode` is never updated, so all extensions are inserted immediately after the viewport node.

This causes the extensions to be written into the markup in the opposite order in which they are detected:

```
C
B
A
```

This seems illogical and might lead to bugs in case ordering would be import.

This PR changes the code so that the `referenceNode` is always moved forward when an extension is added, and therefore keeps the ordering of the extensions intact:
```
A
B
C
```


sebastianbenz (approved):
Nice - thanks!

PR #1145, opened 2021-02-24
🚀 Add icon font linter rule

lluerich:
This checks if commonly used classnames are being used on a page and checks further for known font-families of icon fonts in the ```@font-face``` declarations in the ```<style amp-custom>``` element.


packages/linter/package.json
```diff
@@ -21,6 +21,8 @@
     "cross-fetch": "3.0.6",
     "debug": "4.3.1",
     "execa": "5.0.0",
+    "postcss": "^8.2.6",
```
sebastianbenz, line 24: please pin versions

packages/linter/src/rules/NoIconFontIsUsed.ts
```diff
@@ -0,0 +1,103 @@
...
+            }
+          }
+        });
+      };
+    });
+
+    // grab the amp-custom css
+    const stylesText = $("style[amp-custom]").html();
+
+    if (stylesText) {
```
sebastianbenz, line 78: return early instead

packages/linter/src/rules/NoIconFontIsUsed.ts
```diff
@@ -0,0 +1,103 @@
...
+          console.warn(`Failed to process CSS`, err.message);
+          return { css: stylesText };
+        });
+    }
+
+    if (iconFontMatches.length > 0) {
+      return this.fail("It seems like icon fonts are being used on this page.");
+    }
+
+    return this.warn("Suspicious icon font class names detected.");
```
sebastianbenz, line 94: How likely is this case?

packages/linter/src/rules/NoIconFontIsUsed.ts
```diff
@@ -0,0 +1,103 @@
+import { Context } from "../index";
+import { Rule } from "../rule";
+
+const postcss = require("postcss");
+const safeParser = require("postcss-safe-parser");
+
+const ICON_FONT_IDENTIFIERS = [
+  {
+    className: "fa-",
```
sebastianbenz, line 9: Do we even need to match by classname? Or is it common that icon font names are changed?

sebastianbenz (approved):
Looks good!

packages/linter/src/rules/NoIconFontIsUsed.ts
lluerich: Well so far I know of one case, concerning the Material Icons Font. The stylesheet is loaded via Google Web Fonts, so there is no custom css to scan for font-families, only classnames. So in this case we might give out a warning.

packages/linter/src/rules/NoIconFontIsUsed.ts
lluerich: As mentioned above, there are some cases that use an external stylesheet via Google Web Fonts, and a way to identify those, would be to match by classname.

packages/linter/package.json
lluerich: 👍🏻

packages/linter/src/rules/NoIconFontIsUsed.ts
sebastianbenz: In that case we could also check for `<link href="https://fonts.googleapis.com/icon?family=Material+Icons"
      rel="stylesheet">`

packages/linter/src/rules/NoIconFontIsUsed.ts
```diff
@@ -41,6 +42,28 @@ export class NoIconFontIsUsed extends Rule {
...
+        if (identifier.hasOwnProperty("url") && identifier["url"]) {
+          return (
+            $(`link[rel='stylesheet'][href^='${identifier.url}']`).length > 0
+          );
+        }
+      }
+    );
+
+    if (knownExternalStylesheets.length) {
+      return this.fail("It seems like icon fonts are being used on this page.");
```
sebastianbenz, line 57: Rephrase to 'Avoid using icon fonts to improve loading speed and accessibility'.

sebastianbenz (approved):

PR #1147, opened 2021-02-25
Add amp-bind optimizer

jridgewell:
See https://github.com/ampproject/amphtml/pull/32851

Part of https://github.com/ampproject/amphtml/issues/27590

### DIFF
--- a/packages/optimizer/lib/DomTransformer.js
+++ b/packages/optimizer/lib/DomTransformer.js
@@ -37,6 +37,7 @@ const TRANSFORMATIONS_AMP_FIRST = [
   // Detect hero image and preload link rel=preload, needs to run after OptimizeImages
   'OptimizeHeroImages',
   // Inject a querySelectorAll query-able i-amphtml-binding attribute on elements with bindings.
+  // This needs to run after AutoExtensionImporter.
   'OptimizeAmpBind',
   // Applies server-side-rendering optimizations
   'ServerSideRendering',
@@ -76,6 +77,7 @@ const TRANSFORMATIONS_PAIRED_AMP = [
   // Detect hero image and preload link rel=preload
   'OptimizeHeroImages',
   // Inject a querySelectorAll query-able i-amphtml-binding attribute on elements with bindings.
+  // This needs to run after AutoExtensionImporter.
   'OptimizeAmpBind',
   // Applies server-side-rendering optimizations
   'ServerSideRendering',
--- a/packages/optimizer/lib/DomTransformer.js
+++ b/packages/optimizer/lib/DomTransformer.js
@@ -36,6 +36,8 @@ const TRANSFORMATIONS_AMP_FIRST = [
   'OptimizeImages',
   // Detect hero image and preload link rel=preload, needs to run after OptimizeImages
   'OptimizeHeroImages',
+  // Inject a querySelectorAll query-able i-amphtml-binding attribute on elements with bindings.
+  'OptimizeAmpBind',
   // Applies server-side-rendering optimizations
   'ServerSideRendering',
   // Removes the boilerplate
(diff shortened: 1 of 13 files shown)


packages/optimizer/lib/DomTransformer.js
```diff
@@ -36,6 +36,8 @@ const TRANSFORMATIONS_AMP_FIRST = [
   'OptimizeImages',
   // Detect hero image and preload link rel=preload, needs to run after OptimizeImages
   'OptimizeHeroImages',
+  // Inject a querySelectorAll query-able i-amphtml-binding attribute on elements with bindings.
```
sebastianbenz, line 39:
```suggestion
  // Inject a querySelectorAll query-able i-amphtml-binding attribute on elements with bindings.
  // This needs to run after AutoExtensionImporter.
```

packages/optimizer/lib/DomTransformer.js
```diff
@@ -73,6 +75,8 @@ const TRANSFORMATIONS_PAIRED_AMP = [
   'OptimizeImages',
   // Detect hero image and preload link rel=preload
   'OptimizeHeroImages',
+  // Inject a querySelectorAll query-able i-amphtml-binding attribute on elements with bindings.
```
sebastianbenz, line 78:
```suggestion
  // Inject a querySelectorAll query-able i-amphtml-binding attribute on elements with bindings.
  // This needs to run after AutoExtensionImporter.
```

packages/optimizer/lib/transformers/OptimizeAmpBind.js
```diff
@@ -0,0 +1,79 @@
...
+const {skipNodeAndChildren} = require('../HtmlDomHelper');
+const {isTemplate} = require('../AmpConstants');
+
+/**
+ */
+class OptimizeAmpBind {
+  constructor(config) {
+    this.log_ = config.log.tag('OptimizeAmpBind');
+
+    // setup implementation only if placeholder generation is enabled
```
sebastianbenz, line 27: Looks like a leftover from a different transformer

```suggestion
```

packages/optimizer/lib/transformers/OptimizeAmpBind.js
```diff
@@ -0,0 +1,79 @@
...
+    this.enabled_ = !!config.optimizeAmpBind;
```
sebastianbenz, line 28: We should enable this by default.

```suggestion
    this.enabled_ = config.optimizeAmpBind !== false;
```

sebastianbenz: Or did you intentionally leave this disabled until the validator changes are live?

sebastianbenz (changes requested):
Nice! 

Two comments:

- using `data-amphtml-binding` instead of `I-amphtml-binding` would mean that we wouldn't have to update the validator
- can you please add a test here: https://github.com/ampproject/amp-toolbox/tree/main/packages/optimizer/spec/transformers/valid. You can run `npm run optimier:test:snapshot` to generate the expected output files

sebastianbenz:
//cc @westonruter @schlessera for porting this to the PHP optimizer.

packages/optimizer/lib/transformers/OptimizeAmpBind.js
jridgewell: Yup, this is intentionally disabled until validator changes are live in AMP Cache. I was gonna follow up with a PR to enable it afterwards.

jridgewell (commented):
> - using data-amphtml-binding instead of I-amphtml-binding would mean that we wouldn't have to update the validator

I actually chose not to do this, since the attribute can only be valid in transformed outputs. Data attributes are always valid.

> - can you please add a test here: https://github.com/ampproject/amp-toolbox/tree/main/packages/optimizer/spec/transformers/valid. You can run npm run optimier:test:snapshot to generate the expected output files

Working on that now.

jridgewell:
Tests added

sebastianbenz (approved):
Thanks!

PR #1148, opened 2021-02-25
♻️ Refine icon font detection

lluerich:
* adds selectors
* selector is now case-insensitive
* print warning instead of failure
* update copy to match other warnings


packages/linter/src/rules/NoIconFontIsUsed.ts
```diff
@@ -26,6 +26,14 @@ const ICON_FONT_IDENTIFIERS = [
     className: "ociton-",
```
sebastianbenz, line 26: typo? `ociton`

packages/linter/src/rules/NoIconFontIsUsed.ts
```diff
@@ -26,6 +26,14 @@ const ICON_FONT_IDENTIFIERS = [
...
     fontFamilies: ["Octicons"],
   },
+  {
+    className: "icn-",
+    fontFamilies: ["icon"],
+  },
+  {
+    className: "icon-",
```
sebastianbenz, line 34: This is a super generic name which I wouldn't warn about without the font name.

packages/linter/src/rules/NoIconFontIsUsed.ts
```diff
@@ -35,7 +43,7 @@ export class NoIconFontIsUsed extends Rule {
   run({ $ }: Context) {
     // check for known classnames
     const iconFontCandidates = ICON_FONT_IDENTIFIERS.filter((identifier) => {
-      return $(`[class*=${identifier.className}]`).length > 0;
+      return $(`[class*=${identifier.className} i]`).length > 0;
```
sebastianbenz, line 46: Out of curiosity, what does the `i` do here?

sebastianbenz: case sensitivity?

lluerich: Exactly :)

packages/linter/src/rules/NoIconFontIsUsed.ts
lluerich: Oh sorry, fixed.

sebastianbenz (approved):

PR #1149, opened 2021-02-26
🔧 Add icomoon

lluerich:
* Add ```icons``` and ```icomoon``` font-family


packages/linter/src/rules/NoIconFontIsUsed.ts
```diff
@@ -32,7 +32,7 @@ const ICON_FONT_IDENTIFIERS = [
   },
   {
     className: "icon-",
-    fontFamilies: ["icon"],
+    fontFamilies: ["icon", "icons", "icomoon"],
```
sebastianbenz, line 35: Can you add a test case? Have we already got tests for all the icon font providers?

sebastianbenz (approved):
Thanks!

PR #1151, opened 2021-02-28
Show blurredPlaceholders warning only when enabled

att55:
### Overview

Hi there!

I found that the warning of blurredPlaceholders was shown even if  blurredPlaceholders option was disabled.
https://github.com/ampproject/amp-toolbox/tree/main/packages/optimizer#blurry-image-placeholders

```
AMP Optimizer AddBlurryImagePlaceholders WARNING jimp and lru-cache need to be installed via `npm install jimp lru-cache` for this transformer to work
```

I don't think there is a need to show the waring when the option is disabled.
So, I change that the waring is shown only when the option is enabled.


sebastianbenz (approved):
Good one! Thanks for fixing!

PR #1153, opened 2021-03-02
♻️ Refactor icon font detection

lluerich:
* uses list of grouped parameters
* fixes postcss-8 deprecation warning


sebastianbenz (approved):
Can you add tests for the new logic?

PR #1161, opened 2021-03-16
when using module/nomodule make sure to use `modulepreload` instead of `preload` for the priority hint

erwinmombay:
Currently chrome double downloads the module (mjs) script twice when we use `preload` instead of `modulepreload` in Chrome. Safari does not support `modulepreload` unfortunately but this is the tradeoff we are making


sebastianbenz (approved):
Thanks for the fix! How long does it take for the validator changes to go live?

erwinmombay:
i believe @honeybadgerdontcare is working on the validator changes so he'll have a more accurate timeline than i can estimate.

honeybadgerdontcare:
It's in cl/363067094 and will be live this week or early next.

erwinmombay:
this will be temporary and we will most likely need to revert back to `preload` at some point, should we live this PR as is or have some sort of mechanism in place to toggle this in the optimizer?

honeybadgerdontcare:
Temporary is between now and earliest May. Even then some users may not upgrade to the latest, correct? In that case it may be a bit longer, but I would think this could be reverted instead of toggled when that version of Chrome is widely adopted.

It'll continue to be supported in the validator regardless.

sebastianbenz:
OK, I'll merge this PR once validator changes are live. We can revert behaviour in a subsequent release.

packages/optimizer/spec/transformers/experimental/RewriteAmpUrls/adds_geoapi/expected_output.html
```diff
@@ -9,8 +9,8 @@
   <link rel="stylesheet" href="https://cdn.ampproject.org/v0.css">
   <link href="https://fonts.googleapis.com/css?foobar" rel="stylesheet" type="text/css">
   <link href="https://example.com/favicon.ico" rel="icon">
-  <link as="script" crossorigin="anonymous" href="https://cdn.ampproject.org/v0.mjs" rel="preload">
+  <link as="script" crossorigin="anonymous" href="https://cdn.ampproject.org/v0.mjs" rel="modulepreload">
```
sebastianbenz, line 12: @erwinmombay are `as=script` and `crossorigin: 'anonymous'` required? `as=script` currently gives a validation error.

honeybadgerdontcare: It's not in prod which is why there is an error. See cl/363067094.

honeybadgerdontcare: The next release for JavaScript Validator should be tomorrow. @MichaelRybak please ping this thread when cl/363067094 goes live with the deployment.

jridgewell: Neither is strictly necessary at the browser level. The validator will require both.

sebastianbenz: Thanks for the clarification. Then I was just confused because `rel="modulepreload"` was already supported in the validator.

sebastianbenz:
Just verified in the AMP Validator that it's now valid AMP.

PR #1164, opened 2021-03-18
Font check 2

sebastianbenz:
Improve font check message for pixi integration

* use lighthouse format for adding font details
* disable checks requiring desktop viewport info


sebastianbenz:
Merging so that we can move on.

PR #1165, opened 2021-03-18
Check for latest component versions

lluerich:
This checks whether a page uses the latest version of each component provided by the validator-rules.


sebastianbenz (approved):
Perfect - thank you!

PR #1167, opened 2021-03-19
add check for module use on transformed pages

patrickkettner:


packages/page-experience/lib/PageDataGatherer.js
```diff
@@ -114,6 +114,23 @@ class PageAnalyzer {
         return css;
       };
 
+      /**
```
sebastianbenz, line 117: this should go into linter as it's a static check

patrickkettner: Would that be different from [the current check](https://github.com/ampproject/amp-toolbox/blob/main/packages/linter/src/rules/ModuleRuntimeUsed.ts)?

sebastianbenz: No

PR #1171, opened 2021-03-22
PX: detect hero images

sebastianbenz:
..and suggest using data-hero attribute


sebastianbenz:
Hm. Looks like tests are non-deterministic as they depend on browser output.

packages/page-experience/lib/checks/heroImages.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+const isHeroImageCandidate = require('../helpers/isHeroImageCandidate');
+
+/**
+ * Checks if hero images are present in the initial viewport and declare data-hero
+ *
+ * @param {Object} pageData
+ */
+const earlyIframes = (pageData, result) => {
```
lluerich, line 24: I guess this should be called heroImages or something more descriptive.

packages/page-experience/lib/checks/heroImages.js
```diff
@@ -0,0 +1,45 @@
...
+    (img) => !img.dataHero && isHeroImageCandidate(img)
+  );
+  if (heroImageCandidates.length === 0) {
+    return;
+  }
+
+  result.warn({
+    title: 'Improve LCP by optimizing hero images',
+    message:
+      'Let AMP Caches and Optimizers optimize your hero images, but adding the `data-hero` attribute to images in your first viewport.',
```
lluerich, line 35: *"by adding"

lluerich (approved):
Two little nits and as you already mentioned the the one test relies on browser output and thus fails.

sebastianbenz:
Thanks for the review! Fixed the nits.

lluerich (approved):
LGTM

PR #1174, opened 2021-03-22
PX: enable mobile device emulation

sebastianbenz:


patrickkettner:
what was the failing case? Does the test updated reflect it?

sebastianbenz:
The tests are independent of the actual viewport dimension, but the results for wrong for real webpages.

PR #1184, opened 2021-03-29
Remove hard-coded isTiny() result for intrinsic layout

schlessera:
There is an error in the isTiny() where the width & height are ignored for an intrinsic layout. This is not correct, as for the intrinsic layout, the image is shrinking to fit the container, but can only grow up to the maximum of its width and height attributes. So if these are below the isTiny() threshold, it should be flagged accordingly.

> Related PR in the PHP toolbox: https://github.com/ampproject/amp-toolbox-php/pull/115


sebastianbenz (commented):
Thanks for fixing!

schlessera:
Well, that was fast! Thanks, @sebastianbenz !

PR #1188, opened 2021-04-01
Make ES modules option consistent

schlessera:
ES modules were changed to be used by default in the following PR: https://github.com/ampproject/amp-toolbox/pull/1140

This also seems to have changed the config setting from `experimentEsm` to `esmModulesEnabled`.

However, the `experimentEsm` option was not replaced everywhere, and the class docblock of the `RewriteAmpUrls` transformer even documented both.

The config being set in some of the spec tests is still using `experimentEsm`, even though that has no function anymore. The fact that the tests still pass is only due to the change of the default from `false` to `true`.


sebastianbenz (approved):
Good one! Thanks for making this more consistent!

PR #1192, opened 2021-04-06
Filter duplicate extensions by src URL

### LINKED ISSUE #1191: ReorderHead deduplicates based on name only
The `ReorderHead` transformer uses the extension name to deduplicate scripts for a same extension. This means that if multiple `<script>` elements use the same extension name (like for `.mjs` script + `nomodule` fallback), they are "deduplicated" and one of these gets lost.

The ReorderHead transformer needs a more robust algorithm for deduplication that takes the requirement for two `<script>` elements for ESM into account. Here's what the Go version is doing: https://github.com/ampproject/amppackager/commit/8e2622daa536d5d49dbb6c80f39cbffe65053f72#diff-4e49b75e7da4b38f1a1be19bfc9c6b9e12dbe964eda574cf83ee419776c782f0R80-R94

sebastianbenz:
Fixes #1191

### DIFF
--- a/packages/optimizer/lib/transformers/ReorderHeadTransformer.js
+++ b/packages/optimizer/lib/transformers/ReorderHeadTransformer.js
@@ -50,12 +50,12 @@ class HeadNodes {
   }
 
   _removeDuplicateCustomExtensions(extensions) {
-    const nodesByName = new Map();
+    const nodesBySrc = new Map();
     for (const node of extensions) {
-      const name = this._getName(node);
-      nodesByName.set(name, node);
+      const src = node.attribs['src'];
+      nodesBySrc.set(src, node);
     }
-    return Array.from(nodesByName.values());
+    return Array.from(nodesBySrc.values());
   }
 
   appendToHead(head) {
--- a/packages/optimizer/spec/transformers/valid/ReorderHeadTransformer/supports_esm/expected_output.html
+++ b/packages/optimizer/spec/transformers/valid/ReorderHeadTransformer/supports_esm/expected_output.html
@@ -1,0 +1,16 @@
+<!doctype html>
+<html ⚡>
+<head>
+  <link rel="stylesheet" href="https://cdn.ampproject.org/v0.css">
+  <script async src="https://cdn.ampproject.org/v0.mjs" type="module" crossorigin="anonymous"></script>
+  <script async nomodule src="https://cdn.ampproject.org/v0/amp-experiment-0.1.js" custom-element="amp-experiment"></script>
+  <script async custom-element="amp-experiment" src="https://cdn.ampproject.org/v0/amp-experiment-0.1.mjs" type="module" crossorigin="anonymous"></script>
+  <script async nomodule src="https://cdn.ampproject.org/v0/amp-mustache-0.2.js" custom-template="amp-mustache"></script>
+  <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.2.mjs" type="module" crossorigin="anonymous"></script>
+  <link href="https://example.com/favicon.ico" rel="icon">
+  <link rel="preload" href="https://cdn.ampproject.org/v0.css" as="style">
+  <link href="https://fonts.googleapis.com/css?foobar" rel="stylesheet" type="text/css">
+  <link as="script" crossorigin="anonymous" href="https://cdn.ampproject.org/v0.mjs" rel="modulepreload">
+</head>
+<body></body>
+</html>
--- a/packages/optimizer/spec/transformers/valid/ReorderHeadTransformer/supports_esm/input.html
+++ b/packages/optimizer/spec/transformers/valid/ReorderHeadTransformer/supports_esm/input.html
@@ -1,0 +1,19 @@
+<!doctype html>
+<html ⚡>
+<head>
+  <script async nomodule src="https://cdn.ampproject.org/v0.js"></script>
+  <script async src="https://cdn.ampproject.org/v0.mjs" type="module" crossorigin="anonymous"></script>
+  <script async nomodule src="https://cdn.ampproject.org/v0/amp-mustache-0.2.js" custom-template="amp-mustache"></script>
+  <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.2.mjs" type="module" crossorigin="anonymous"></script>
+  <script async nomodule src="https://cdn.ampproject.org/v0/amp-experiment-0.1.js" custom-element="amp-experiment"></script>
+  <script async nomodule src="https://cdn.ampproject.org/v0/amp-experiment-0.1.js" custom-element="amp-experiment"></script>
+  <script async custom-element="amp-experiment" src="https://cdn.ampproject.org/v0/amp-experiment-0.1.mjs" type="module" crossorigin="anonymous"></script>
+  <script async custom-element="amp-experiment" src="https://cdn.ampproject.org/v0/amp-experiment-0.1.mjs" type="module" crossorigin="anonymous"></script>
+  <link rel="stylesheet" href="https://cdn.ampproject.org/v0.css">
+  <link href="https://fonts.googleapis.com/css?foobar" rel="stylesheet" type="text/css">
+  <link href="https://example.com/favicon.ico" rel="icon">
+  <link as="script" crossorigin="anonymous" href="https://cdn.ampproject.org/v0.mjs" rel="modulepreload">
+  <link rel="preload" href="https://cdn.ampproject.org/v0.css" as="style">
+</head>
+<body></body>
+</html>


westonruter:
Testing PHP implementation: https://github.com/ampproject/amp-toolbox-php/pull/130

packages/optimizer/spec/transformers/valid/ReorderHeadTransformer/supports_esm/expected_output.html
```diff
@@ -0,0 +1,16 @@
...
+  <link rel="stylesheet" href="https://cdn.ampproject.org/v0.css">
+  <script async src="https://cdn.ampproject.org/v0.mjs" type="module" crossorigin="anonymous"></script>
+  <script async nomodule src="https://cdn.ampproject.org/v0/amp-experiment-0.1.js" custom-element="amp-experiment"></script>
+  <script async custom-element="amp-experiment" src="https://cdn.ampproject.org/v0/amp-experiment-0.1.mjs" type="module" crossorigin="anonymous"></script>
+  <script async nomodule src="https://cdn.ampproject.org/v0/amp-mustache-0.2.js" custom-template="amp-mustache"></script>
+  <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.2.mjs" type="module" crossorigin="anonymous"></script>
+  <link href="https://example.com/favicon.ico" rel="icon">
+  <link rel="preload" href="https://cdn.ampproject.org/v0.css" as="style">
+  <link href="https://fonts.googleapis.com/css?foobar" rel="stylesheet" type="text/css">
+  <link as="script" crossorigin="anonymous" href="https://cdn.ampproject.org/v0.mjs" rel="modulepreload">
```
westonruter, line 13: Shouldn't the `link[rel=preload]` and `link[rel=modulepreload]` be moved to the beginning of the `head`?

sebastianbenz: Done

westonruter:
Here's the current order in the PHP implementation:

```html
  <link as="script" crossorigin="anonymous" href="https://cdn.ampproject.org/v0.mjs" rel="modulepreload">
  <link rel="preload" href="https://cdn.ampproject.org/v0.css" as="style">
  <link rel="stylesheet" href="https://cdn.ampproject.org/v0.css">
  <script async nomodule src="https://cdn.ampproject.org/v0.js"></script>
  <script async src="https://cdn.ampproject.org/v0.mjs" type="module" crossorigin="anonymous"></script>
  <script async nomodule src="https://cdn.ampproject.org/v0/amp-experiment-0.1.js" custom-element="amp-experiment"></script>
  <script async custom-element="amp-experiment" src="https://cdn.ampproject.org/v0/amp-experiment-0.1.mjs" type="module" crossorigin="anonymous"></script>
  <script async="" nomodule="" src="https://cdn.ampproject.org/v0/amp-mustache-0.2.js" custom-template="amp-mustache"></script>
  <script async="" custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.2.mjs" type="module" crossorigin="anonymous"></script>
  <link href="https://example.com/favicon.ico" rel="icon">
  <link href="https://fonts.googleapis.com/css?foobar" rel="stylesheet" type="text/css">
```

When outputting AMP scripts, it seems that the `module`/`nomodule` versions should be next to each other in general, right? Also, it seems the `module` version should always come first? That isn't yet the case for the PHP transformer.

sebastianbenz:
> When outputting AMP scripts, it seems that the module/nomodule versions should be next to each other in general, right? Also, it seems the module version should always come first? That isn't yet the case for the PHP transformer.

Not sure if that really makes a difference. I'd assume that the browser will simply ignore the `nomodule` tag and move on.

packages/optimizer/spec/transformers/valid/ReorderHeadTransformer/supports_esm/expected_output.html
```diff
@@ -0,0 +1,16 @@
...
+  <link rel="preload" href="https://cdn.ampproject.org/v0.css" as="style">
+  <link rel="stylesheet" href="https://cdn.ampproject.org/v0.css">
+  <script async src="https://cdn.ampproject.org/v0.mjs" type="module" crossorigin="anonymous"></script>
+  <script async nomodule src="https://cdn.ampproject.org/v0/amp-experiment-0.1.js" custom-element="amp-experiment"></script>
+  <script async custom-element="amp-experiment" src="https://cdn.ampproject.org/v0/amp-experiment-0.1.mjs" type="module" crossorigin="anonymous"></script>
+  <script async nomodule src="https://cdn.ampproject.org/v0/amp-mustache-0.2.js" custom-template="amp-mustache"></script>
+  <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.2.mjs" type="module" crossorigin="anonymous"></script>
+  <link href="https://example.com/favicon.ico" rel="icon">
+  <link href="https://fonts.googleapis.com/css?foobar" rel="stylesheet" type="text/css">
+  <link as="script" crossorigin="anonymous" href="https://cdn.ampproject.org/v0.mjs" rel="modulepreload">
```
westonruter, line 13: This needs to be moved up along with the `preload` as well. Ref https://github.com/ampproject/amp-toolbox-php/pull/130/commits/a3a99fd5c66dad645a1ab47138d000b32b635df4

sebastianbenz: Thanks for the patience taking it up with my empty brain today...

packages/optimizer/spec/transformers/valid/ReorderHeadTransformer/supports_esm/input.html
```diff
@@ -0,0 +1,19 @@
+<!doctype html>
+<html ⚡>
+<head>
+  <script async nomodule src="https://cdn.ampproject.org/v0.js"></script>
```
westonruter, line 4: Big problem: The `nomodule` v0.js script is getting dropped in the output. Here's one thing I had to do in the PHP version to account for this: https://github.com/ampproject/amp-toolbox-php/pull/130/commits/4bad718a538c317168a9cb36728a7d977e25b7db

packages/optimizer/spec/transformers/valid/ReorderHeadTransformer/supports_esm/expected_output.html
```diff
@@ -0,0 +1,16 @@
+<!doctype html>
+<html ⚡>
+<head>
+  <link as="script" crossorigin="anonymous" href="https://cdn.ampproject.org/v0.mjs" rel="modulepreload">
+  <link rel="preload" href="https://cdn.ampproject.org/v0.css" as="style">
+  <link rel="stylesheet" href="https://cdn.ampproject.org/v0.css">
+  <script async src="https://cdn.ampproject.org/v0.mjs" type="module" crossorigin="anonymous"></script>
```
westonruter, line 7:
```suggestion
  <script async src="https://cdn.ampproject.org/v0.mjs" type="module" crossorigin="anonymous"></script>
  <script async nomodule src="https://cdn.ampproject.org/v0.js"></script>
```

westonruter:
> Not sure if that really makes a difference. I'd assume that the browser will simply ignore the `nomodule` tag and move on.

Can we go ahead and do that so that there is consistency in the ordering? I've implemented this in the PHP version now. The PHP version now generates the `head` as follows for `ReorderHeadTransformer/supports_esm/input.html`:

```html
  <link as="script" crossorigin="anonymous" href="https://cdn.ampproject.org/v0.mjs" rel="modulepreload">
  <link rel="preload" href="https://cdn.ampproject.org/v0.css" as="style">
  <link rel="stylesheet" href="https://cdn.ampproject.org/v0.css">
  <script async src="https://cdn.ampproject.org/v0.mjs" type="module" crossorigin="anonymous"></script>
  <script async nomodule src="https://cdn.ampproject.org/v0.js"></script>
  <script async custom-element="amp-experiment" src="https://cdn.ampproject.org/v0/amp-experiment-0.1.mjs" type="module" crossorigin="anonymous"></script>
  <script async nomodule src="https://cdn.ampproject.org/v0/amp-experiment-0.1.js" custom-element="amp-experiment"></script>
  <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.2.mjs" type="module" crossorigin="anonymous"></script>
  <script async nomodule src="https://cdn.ampproject.org/v0/amp-mustache-0.2.js" custom-template="amp-mustache"></script>
  <link href="https://example.com/favicon.ico" rel="icon">
  <link href="https://fonts.googleapis.com/css?foobar" rel="stylesheet" type="text/css">
```

This ordering aligns with the AMP boilerplate usage as well, where we have `style[amp-boilerplate]` followed by `noscript > noscript[amp-boilerplate]`. In general it seems preferable for the thing which is the normal case to be first. For extremely slow connections, the module version could be loaded sooner than the non-module version if the HTML loading gets interrupted.

westonruter:
The PHP version is also sorting the script components by name, so `amp-experiment` script(s) will come before `amp-mustache` for example.

packages/optimizer/lib/transformers/ReorderHeadTransformer.js
```diff
@@ -50,24 +50,24 @@ class HeadNodes {
...
+    const nodesBySrc = new Map();
     for (const node of extensions) {
-      const name = this._getName(node);
-      nodesByName.set(name, node);
+      const src = node.attribs['src'];
+      nodesBySrc.set(src, node);
     }
-    return Array.from(nodesByName.values());
+    return Array.from(nodesBySrc.values());
   }
```
westonruter, line 59: I took an alternative approach to this in PHP, to do the deduplication in the `registerScript` function itself: https://github.com/ampproject/amp-toolbox-php/pull/130/files#diff-b0da2cc892f74bf1ebc7a84e63d7065894b24541a48f3a26089f23860ee1209fR156-R190

This also accounts for `module` and `nomodule` variations.

packages/optimizer/lib/transformers/ReorderHeadTransformer.js
```diff
@@ -50,24 +50,24 @@ class HeadNodes {
...
 
   _removeDuplicateCustomExtensions(extensions) {
-    const nodesByName = new Map();
+    const nodesBySrc = new Map();
     for (const node of extensions) {
-      const name = this._getName(node);
-      nodesByName.set(name, node);
+      const src = node.attribs['src'];
```
schlessera, line 55: Shouldn't this remove the host to make sure a same extension is not downloaded twice from two different hosts?

packages/optimizer/spec/transformers/valid/ReorderHeadTransformer/supports_esm/expected_output.html
```diff
@@ -0,0 +1,17 @@
...
+<head>
+  <link as="script" crossorigin="anonymous" href="https://cdn.ampproject.org/v0.mjs" rel="modulepreload">
+  <link rel="preload" href="https://cdn.ampproject.org/v0.css" as="style">
+  <link rel="stylesheet" href="https://cdn.ampproject.org/v0.css">
+  <script async nomodule src="https://cdn.ampproject.org/v0.js"></script>
+  <script async src="https://cdn.ampproject.org/v0.mjs" type="module" crossorigin="anonymous"></script>
+  <script async nomodule src="https://cdn.ampproject.org/v0/amp-experiment-0.1.js" custom-element="amp-experiment"></script>
+  <script async custom-element="amp-experiment" src="https://cdn.ampproject.org/v0/amp-experiment-0.1.mjs" type="module" crossorigin="anonymous"></script>
+  <script async nomodule src="https://cdn.ampproject.org/v0/amp-mustache-0.2.js" custom-template="amp-mustache"></script>
+  <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.2.mjs" type="module" crossorigin="anonymous"></script>
```
westonruter, line 12: Can the order be reversed, with the `nomodule` one coming after?

packages/optimizer/lib/transformers/ReorderHeadTransformer.js
```diff
@@ -100,31 +96,39 @@ class HeadNodes {
   }
 
   _registerScript(node) {
+    const moduleIndex = hasAttribute(node, 'nomodule') ? 0 : 1;
```
westonruter, line 99: This should ensure `nomodule` comes after `type=module`.

```suggestion
    const moduleIndex = hasAttribute(node, 'nomodule') ? 1 : 0;
```

packages/optimizer/spec/transformers/valid/ReorderHeadTransformer/supports_esm/expected_output.html
sebastianbenz: done

packages/optimizer/spec/transformers/valid/ReorderHeadTransformer/supports_esm/input.html
sebastianbenz: fixed

sebastianbenz (commented):
Addressed the feedback. Scripts should now have the right order. 

Not super happy about the changes as the implementation is now more complex (and expensive) than before (without a real need, as the module scripts are injected in a later stage of the transformation).

packages/optimizer/lib/transformers/ReorderHeadTransformer.js
```diff
@@ -41,33 +41,30 @@ class HeadNodes {
...
+  _sortExtensions(extensions) {
+    const sortedExtensions = new Map([...extensions].sort((a, b) => a[0].localeCompare(b[0])));
+    // TODO replace with Array#flat once Node 10 is EOL
+    return [].concat.apply([], Array.from(sortedExtensions.values()));
   }
 
   appendToHead(head) {
     appendChild(head, this._metaCharset);
+    appendAll(head, this._metaOther);
+    appendAll(head, this._resourceHintLinks);
```
schlessera, line 61: The `metaOther` can potentially includes lots of non-critical stuff like OpenGraph tags. I think in terms of performance, the resource hints should come right after the charset and before these other meta tags.

sebastianbenz: good point! done

packages/optimizer/lib/DomTransformer.js
```diff
@@ -44,16 +44,16 @@ const TRANSFORMATIONS_AMP_FIRST = [
   // Removes the boilerplate
   // needs to run after ServerSideRendering
   'AmpBoilerplateTransformer',
-  // Optimizes script import order
-  // needs to run after ServerSideRendering
-  'ReorderHeadTransformer',
   // needs to run after ReorderHeadTransformer
   'RewriteAmpUrls',
```
schlessera, line 48:
```suggestion
  'RewriteAmpUrls',
```

packages/optimizer/lib/DomTransformer.js
```diff
@@ -86,16 +86,16 @@ const TRANSFORMATIONS_PAIRED_AMP = [
...
```
schlessera, line 90:
```suggestion
  'RewriteAmpUrls',
```

packages/optimizer/lib/DomTransformer.js
```diff
@@ -119,13 +119,13 @@ const TRANSFORMATIONS_MINIMAL = [
...
```
schlessera, line 123:
```suggestion
  'RewriteAmpUrls',
```

schlessera (changes requested):
I'm still wondering about the following priorities:
1. ) Is the viewport needed for making decisions about the resource hints? I don't think so, but I'm not entirely sure.
2. ) The boilerplate style is now much further down, after the scripts. Are we sure this does not lead to FOUC?

sebastianbenz:
re 1) Yes, viewport needs to come first (I think that was the reason why I added `metaOthers` before the resource hints originally. The correct way is to also special case viewport. 
re 2) I'm not sure about this. My thinking for the current order is: component downloads should start asap hence they should be placed before custom css. Boilerplate should come after custom CSS to avoid custom CSS accidentally overriding the boilerplate. FOUC should not be a problem as the browser will parse all CSS before rendering the page.

sebastianbenz:
Hm. Following that logic, we should also move `v0.css` down.

westonruter (approved):
LGTM, but I defer to @schlessera.

schlessera (approved):

schlessera:
:shipit:

sebastianbenz:
Thanks a lot for the patience and detailed review!

PR #1194, opened 2021-04-09
Add crossorigin=anonymous to nomodule scripts the same as module ones

westonruter:
When reviewing the output of the `RewriteAmpUrls` transformer, I noticed that the `nomodule` scripts lacked the `crossorigin=anonymous` attribute like the the `module` scripts did. This makes them consistent.

Aside: It seems to be that there should also be a transformer config option for whether `crossorigin=anonymous` should be added in the first place. Sure it should be added by default, but for some use cases it is not good to add, namely when you are wanting to opt-in to beta/experimental channels via cookie (cf. https://github.com/ampproject/amp-wp/issues/4600).  Nevertheless, for this to be feasible the validator would need to make `crossorigin=anonymous` non-mandatory for the [`modulepreload` link](https://github.com/ampproject/amphtml/blob/0320666d03db37c4e0e363b55f828aa807ae321c/validator/validator-main.protoascii#L303-L307) and all [module scripts](https://github.com/ampproject/amphtml/blob/0320666d03db37c4e0e363b55f828aa807ae321c/validator/validator-main.protoascii#L3076-L3079). I assume the cookie-based opt-in works for modules.


sebastianbenz (approved):
Thanks!

sebastianbenz:
I replicated the behaviour of the AMP cache here. Not sure what the original intention was to omit it for nomodule.

PR #1195, opened 2021-04-09
Add AmpBoilerplateErrorHandler transformer

### LINKED ISSUE #1146: Add support for script amp-onerror
See https://github.com/ampproject/amphtml/issues/22543#issuecomment-673603872.

Support adding `<script amp-onerror>document.querySelector("script[src*='/v0.js']").onerror=function(){document.querySelector('style[amp-boilerplate]').textContent=''}</script>` to the `<head>` in AMP Optimizer.

schlessera:
This PR adds a transformer that adds a `<script amp-onerror>` element to the document head to disable the boilerplate early on errors while loading the AMP runtime.

Fixes #1146


schlessera:
@sebastianbenz
There is a remaining issue because of the whitespace that is added automatically by the JS DOM when adding the script content. This causes the resulting document to fail validation, as the validation rule for this `<script amp-onerror>` element is extremely strict.

I don't know enough about the JS DOM abstraction to solve that. Any ideas on how to ensure no whitespace is added?

sebastianbenz:
The transformer is fine, we run a code formatter after the transformations (for readability) which adds the whitespace. You have disable this, by adding `script` here: https://github.com/ampproject/amp-toolbox/blob/882c791c8c27d0dfdb947a90932913bbc8f9c129/packages/optimizer/spec/helpers/TransformerRunner.js#L25

You'll also have to run `npm run test:snapshot:optimizer` again.

schlessera:
The CLA check seems to be stuck here...

schlessera:
When disabling the beautify on scripts, it not only leaves the internals of each script tag intact, but it also removes the newlines between the individual elements. Is there a way to avoid styling the script internals, but keep the newlines outside of the script element intact?

sebastianbenz:
Not sure if that's configurable. I don't consider this a critical problem.

sebastianbenz (approved):

PR #1196, opened 2021-04-09
Prevent SSR for amp-img occurring after second paragraph

westonruter:
As pointed out by @jono-alderson, hero image prerendering can be detrimental when there are no hero images in the first viewport, as seen in these two examples from amp-wp.org:

[Blog post](https://amp-wp.org/building-amp-first-websites-with-wordpress-a-roadmap/) | [Homepage](https://amp-wp.org/)
-------------------------------------------------------------------------------------------|----------------------
image | image

This is especially problematic because `OptimizeHeroImage` does not include `loading=lazy` on the SSR'ed image. The result is that an optimized page may actually perform _worse_. 

This PR mitigates this issue by only doing hero image prerendering for images that appear before the second paragraph in the page.

Another mitigation measure is to add `loading=lazy` to any SSR'ed hero image that was automatically detected. If a document had `data-hero` on an image up front then the `loading=lazy` would remain omitted since the author is explicitly assuring the optimizer that the image should indeed be in the first viewport.

### DIFF
--- a/packages/optimizer/lib/transformers/OptimizeHeroImages.js
+++ b/packages/optimizer/lib/transformers/OptimizeHeroImages.js
@@ -145,12 +145,17 @@ class OptimizeHeroImage {
     let heroImageCandidate = null;
     let heroImages = [];
     let node = root;
+    let seenParagraphCount = 0;
     // Walk over all nodes in the body
     while (node !== null) {
+      if (node.tagName === 'p') {
+        seenParagraphCount++;
+      }
       // Look for data-hero attribute
       this.addImageWithDataHero(node, heroImages);
-      // Auto detect a hero image in case data-hero is not used
-      if (!heroImageCandidate && heroImages.length === 0) {
+      // Auto detect a hero image in case data-hero is not used,
+      // but only if before the second paragraph.
+      if (!heroImageCandidate && seenParagraphCount < 2 && heroImages.length === 0) {
         heroImageCandidate = this.isCandidateHeroImage(node);
       }
       if (isTemplate(node)) {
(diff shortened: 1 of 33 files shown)


packages/optimizer/lib/transformers/OptimizeHeroImages.js
```diff
@@ -317,12 +322,21 @@ class OptimizeHeroImage {
...
       class: 'i-amphtml-fill-content i-amphtml-replaced-content',
       decoding: 'async',
     });
+
+    // If the image was detected as hero image candidate (and thus lacks an explicit
+    // data-hero), mark it as a hero and add loading=lazy to guard against making
+    // the page performance even worse by eagerly loading an image outside the viewport.
+    if (!('data-hero' in node.attribs)) {
+      node.attribs['data-hero'] = '';
+      imgNode.attribs['loading'] = 'lazy';
```
sebastianbenz, line 337: I really like this approach. Have you checked if it is valid AMP?

westonruter: Yes, it is. The `loading` attribute is allowed, and it can be either `lazy` or `eager`: https://github.com/ampproject/amphtml/blob/a8889ac1b2622c461422a2186366f7a433848f44/validator/validator-main.protoascii#L3834-L3838

sebastianbenz: Seems to be, otherwise the end-to-end spec would fail.

sebastianbenz (approved):
This is great! Thanks Weston!

PR #1198, opened 2021-04-09
Improve logic to detect explicitly marked hero images

schlessera:
The logic in https://github.com/ampproject/amp-toolbox/pull/1196 fails to properly detect a nested construct, like if an iframe if marked with `data-hero` and its placeholder image is SSRed.

This PR improves the logic to properly detect these cases.

### DIFF
--- a/packages/optimizer/lib/transformers/OptimizeHeroImages.js
+++ b/packages/optimizer/lib/transformers/OptimizeHeroImages.js
@@ -20,6 +20,7 @@ const {
   appendChild,
   createElement,
   hasAttribute,
+  remove,
   insertAfter,
   nextNode,
   firstChildByTag,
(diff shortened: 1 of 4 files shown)


packages/optimizer/lib/transformers/OptimizeHeroImages.js
```diff
@@ -367,6 +370,25 @@ class OptimizeHeroImage {
     }
     appendChild(node, imgNode);
   }
+
+  isMarkedAsHeroImage(node) {
+    while (node) {
+      if (!node.tagName) {
```
westonruter, line 376: Why would `tagName` ever be empty?

packages/optimizer/lib/transformers/OptimizeHeroImages.js
```diff
@@ -367,6 +370,25 @@ class OptimizeHeroImage {
...
+        node = node.parent;
```
westonruter, line 377: What if `node.parent` here is empty?

packages/optimizer/lib/transformers/OptimizeHeroImages.js
schlessera: If it's not an element, but a different type of node. I initially tried to check for `HTMLElement`, but that didn't work as I expected.

packages/optimizer/lib/transformers/OptimizeHeroImages.js
schlessera: Oh, it is missing a continue. I'll add it.

packages/optimizer/lib/transformers/OptimizeHeroImages.js
westonruter: It seems elements are only ever passed into this function, but I guess this is good for hardening.

westonruter (approved):

packages/optimizer/lib/transformers/OptimizeHeroImages.js
schlessera: Yes, but this check is within a loop that loops from parent to parent, so this does not only check what is passed into the function, but every ancestor of it as well. And as elements can also be children of non-elements (see https://www.w3schools.com/jsref/prop_node_nodetype.asp), I think this is needed for some edge cases.

westonruter:
@schlessera @sebastianbenz I amended the PR with 16a1a85 which I've done in PHP via https://github.com/ampproject/amp-toolbox-php/pull/149.

sebastianbenz (approved):

PR #1208, opened 2021-04-13
Make origin fetch parallel to the KV check

samouri:
**summary**

The fetch to origin can be launched in parallel to the check against KV store.


powerivq:
Batch closing old PRs that are not worked on.

PR #1209, opened 2021-04-13
cloudflare-optimizer-scripts: optimize images using squoosh

samouri:
**summary**
- For `the_scenic` whose images were already optimized, I reduced max width of images to 1000px and use mozjpeg with quality 75 to bring down images from ~300kb --> 100kb.
- The `ecommerce` template's images had never been optimized, so simply running through mozjpeg often reduced size by > 50%.
- I also fixed the "api" calls in the ecommerce template. Specifically all of the links to images were incorrect.
- As a bonus, added `.DS_Store` to .gitignore

Speed diff accordint to WPT with 9 runs:
- Optimized doc's LCP [0.912s](https://www.webpagetest.org/result/210413_AiDcAX_b46c1a7d8ccfb4354a8e73b528fd4374) --> [0.796s](https://www.webpagetest.org/result/210413_AiDcC9_f7a963022d63d6955205b1a0c7aac8dc/)

cc @sebastianbenz


samouri:
This is another case where I don't trust the CI's cache.

PR #1217, opened 2021-04-22
Enable OptimizeAmpBind transformer by default

schlessera:
As discussed in https://github.com/ampproject/amp-toolbox/pull/1147/files#r582707679 and given the required changes in the validator and runtime are now live, the `OptimizeAmpBind` transformer should be enabled by default once all blockers are resolved.

This does that.


sebastianbenz (approved):
thanks!

PR #1218, opened 2021-04-22
Only recommend Font Display Optional for non-icon fonts

sebastianbenz:
Fixes https://github.com/ampproject/amp.dev/issues/5698


packages/page-experience/lib/checks/fontDisplay.js
```diff
@@ -14,6 +14,19 @@
...
+const ICON_FONTS = new Set([
+  'FontAwesome',
+  'Font Awesome',
+  'NerdFontsSymbols',
+  'Nerd Font',
+  'Material Icons',
+  'IcoFont',
+  'icon',
+  'icons',
+  'icomoon',
```
westonruter, line 27: There are a couple other icon fonts that are common in WordPress:

```suggestion
  'icomoon',
  'Dashicons',
  'Genericons',
```

patrickkettner (approved):

packages/page-experience/lib/checks/fontDisplay.js
```diff
@@ -30,6 +45,11 @@ const checkDisplayOptional = (pageData, result) => {
     }
     const isGoogleFont =
       fontface.mainSrc && fontface.mainSrc.startsWith('https://fonts.gstatic.com');
+    const isIconFont = ICON_FONTS.has(fontface.fontFamily);
```
lluerich, line 48: This will not catch the aforementioned font family ```Font Awesome 5``` unless it is added to the set. Another way would be to check if a substring of the font family is found in the array like:
```
const isIconFont = ICON_FONTS.some((font) => {
  return fontface['fontFamily'].includes(font)
});
```

lluerich (approved):
Nice work! One comment, but LGTM.

packages/page-experience/lib/checks/fontDisplay.js
sebastianbenz: Added these to the list

packages/page-experience/lib/checks/fontDisplay.js
sebastianbenz: good tip! added!

PR #1219, opened 2021-04-24
CF: Use the original request to pass through headers to origin request.

evanderkoogh:
This change is to pass in the original `request` object to the original fetch. This is so that headers that are present on the original request are passed through to the origin request.
Cloudflare adds a headers that are often used by origin request, so this is needed.

### DIFF
--- a/packages/cloudflare-optimizer-scripts/src/index.js
+++ b/packages/cloudflare-optimizer-scripts/src/index.js
@@ -85,7 +85,7 @@ async function handleRequest(event, config) {
     }
   }
 
-  const response = await fetch(url.href, request);
+  const response = await fetch(url, request);
   const clonedResponse = response.clone();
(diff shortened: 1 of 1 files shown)


packages/cloudflare-optimizer-scripts/src/index.js
```diff
@@ -85,9 +85,7 @@ async function handleRequest(event, config) {
     }
   }
 
-  const response = await fetch(url.toString(), {
-    cf: {minify: {html: true}},
-  });
+  const response = await fetch(url.href, request);
```
xtuc, line 88:
```suggestion
  const response = await fetch(url, request);
```

samouri:
@evanderkoogh: if you `@` me on future PRs I'll be able to get to them quicker :)

PR #1205, opened 2021-04-12
[breaking] Add fast and full default configs

sebastianbenz:
Add two new standard configurations optimizing for either performance or performance optimization. This also improve the minimal bundle install size as all dependencies required by non essential transformation steps are now optional.

PR #1220, opened 2021-04-24
Enable optional transformers for their own tests

schlessera:
The [PR for adding fast and full default configs](https://github.com/ampproject/amp-toolbox/pull/1205) disabled some transformers by default and had the spec tests be adapted without running the `AutoExtensionsImporter` and `SeparateKeyframes` transformers for their own respective tests.

This PR adds configuration to enable these transformers for their own tests.

TODO: What about the end-to-end tests? Those we changed considerably with the above PR as well...


sebastianbenz (approved):
Oh yes - my fault! Thanks for fixing this!

sebastianbenz:
End-to-end tests are handled as they're now run for both configurations `fast` and `full`.

PR #1222, opened 2021-04-26
Make cache configurable

sebastianbenz:
The cache implementation can now be customized via the config object. 

`config.cache = myCache` specifies the catch implementation to use for caching artifacts required during transformation (runtime CSS, validation rules, latest runtime version,...). A cache needs to conform to the following interface:

```typescript
interface Cache {
  set(key: Object, value: Object): void;
  get(key: Object): Object;
}
```

By default, artifacts will be cached in a temporary directory. If there is no filesystem access available, the cache will fallback to an in memory implementation.

This PR also fixes a few naming inconsistencies in the readme and tests.


sebastianbenz:
Test failures are unrelated to this PR>

packages/optimizer/README.md
```diff
@@ -205,6 +222,13 @@ const ampOptimizer = AmpOptimizer.create({
 - default: `{}`
 - used by: [AutoExtensionImport](lib/transformers/AutoExtensionImporter.js)
 
+#### `fetch`
+
+Specifies the fetch implementation to use:
```
patrickkettner, line 227: nit, i'd link to something about what `fetch` is (the [spec](https://fetch.spec.whatwg.org/) maybe?)

packages/optimizer/lib/DomTransformer.js
```diff
@@ -20,6 +20,8 @@ const log = require('./log');
 let fetch = require('node-fetch');
 const RuntimeVersion = require('@ampproject/toolbox-runtime-version/lib/RuntimeVersion');
 const fetchRuntimeParameters = require('./fetchRuntimeParameters');
+const cache = require('./cache.js');
+const {OperationCanceledException} = require('typescript');
```
patrickkettner, line 24: is `OperationCanceledException` unused? If so, shouldn't this have been caught by prettier or something?

patrickkettner (approved):
nits and a typo, otherwise LGTM

packages/optimizer/README.md
```diff
@@ -170,6 +170,23 @@ Automatically import any missing AMP Extensions (e.g. amp-carousel). This is not
...
 
+#### `cache`
+
+Specifies the catch implementation to use for caching artifacts required during transformation (runtime CSS, validation rules, latest runtime version,...). A cache needs to conform to the following interface:
+
+```typescript
+interface Cache {
+  set(key: Object, value: Object): void;
+  get(key: Object): Object;
+}
```
schlessera, line 181: This interface guarantees you'll always get an `Object` back.

What about the case where you want to retrieve something from cache that was not yet cached or can't be fetched?

I usually see two different conventions:
- The cache can return `null`, `false` or similar to denote that nothing could be retrieved. This goes counter to strong typing.
- The cache throws an error/exception if nothing could be retrieved. This still enforces strong typing. However, in that case, you usually have a `has()` method as well to be able to avoid the error condition upfront.

If no `has()` is provided, there should be some documentation (docblock?) that defines how the error condition can be detected.

patrickkettner:
Would a promise based interface that resolves to an object be too
cumbersome?--
patrick

schlessera (approved):
LGTM apart from the note about the error condition handling above.

sebastianbenz:
Addressed the comments! Thanks for the review. Good catch with the promises! The actual implementation was already promise based => updated the readme.

PR #1227, opened 2021-04-28
CI: switch to pull_request from pull_request_target

samouri:
**summary**
Having the CI action execute on the `pull_request_target` hook was resulting in the tested SHA to [always be main](https://github.com/actions/checkout/issues/486#issue-869251025). By switching to `pull_request`, the correct merge commit of the PR's branch is now chosen instead. You can confirm this by checking the logs of the GH Action before/after.

**fixed screenshot**
Screen Shot 2021-04-28 at 10 03 02 AM

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [pull_request_target, push]
+on: [pull_request, push]
 
 jobs:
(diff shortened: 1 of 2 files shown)


sebastianbenz (approved):
Hah! That makes so much sense in hindsight! Thanks!

PR #1228, opened 2021-04-28
Simplify CI

sebastianbenz:
- clean should no longer be needed, as caching was not the problem
  causing the CI to fail.
- don't run build twice as it's already run as part of test


samouri (approved):
Love it

PR #1231, opened 2021-05-04
Bundle runtime css and version

sebastianbenz:
This adds offline support to optimizer by downloading the latest runtime
and styles as part of the build step. Transformations will fallback to
these if the latest runtime version and styles cannot be downloaded.

### DIFF
--- a/packages/optimizer/lib/DomTransformer.js
+++ b/packages/optimizer/lib/DomTransformer.js
@@ -124,7 +124,6 @@ const CONFIG_DEFAULT = {
   optimizeAmpBind: true,
   optimizeHeroImages: true,
   separateKeyframes: true,
-  config: 'custom',
 };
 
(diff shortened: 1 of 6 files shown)


packages/optimizer/lib/DomTransformer.js
```diff
@@ -124,6 +124,7 @@ const CONFIG_DEFAULT = {
   optimizeAmpBind: true,
   optimizeHeroImages: true,
   separateKeyframes: true,
+  config: 'custom',
```
patrickkettner, line 127: You have to explicitly set the fact that it's a custom config?￼

patrickkettner (approved):

packages/optimizer/lib/DomTransformer.js
sebastianbenz: This is actually not needed anymore. I had the idea to make caching depend on the config, but decided to keep it simple for now. Removed.

PR #1232, opened 2021-05-04
Add SSR support for fluid layout

westonruter:
The use of the `fluid` layout currently blocks server-side rendering since it is not among the `SUPPORTED_LAYOUTS`:

https://github.com/ampproject/amp-toolbox/blob/3ec7bd40f34adbc8be66f951edfa6d91d5794c01/packages/optimizer/lib/transformers/ApplyLayout.js#L30-L40

The `fluid` layout is not ideal because it inherently involves a layout shift since its dimensions get determined at runtime, and it will not get rendered when it is initially in the viewport (per [docs](https://github.com/ampproject/amphtml/blob/main/extensions/amp-ad-network-doubleclick-impl/fluid.md)). Nevertheless, when this layout is required it would be preferable for it to not prevent other parts of the page from benefiting from SSR. 

Implementation provided by @jridgewell. 

Merging this PR is blocked by the AMP validator being updated, specifically to modify [`validateSsrLayout()`](https://github.com/ampproject/amphtml/blob/f2a49b8efa4ecb0cd1dd273977799299a9e1f816/validator/js/engine/validator.js#L4436-L4441) as follows:

```diff
      const validInternalClasses = Object.create(null);
      validInternalClasses[getLayoutClass(layout)] = 0;
      if (isLayoutSizeDefined(layout)) {
        // i-amphtml-layout-size-defined
        validInternalClasses[getLayoutSizeDefinedClass()] = 0;
      }
+     if ('fluid' === layout) {
+       validInternalClasses['i-amphtml-layout-awaiting-size'] = 0;
+     }
```

- [x] Add tests.
- [x] Verify validator updated and deployed.

cf. b/184254585

### DIFF
--- a/packages/optimizer/lib/transformers/ApplyLayout.js
+++ b/packages/optimizer/lib/transformers/ApplyLayout.js
@@ -81,9 +81,6 @@ function apply(layout, width, height, node) {
       // Do nothing here.
       break;
     case 'fluid':
-      if (width.isSet) {
-        styles = `width:${width.numeral}${width.unit};`;
-      }
       styles += 'height:0;';
       addClass(node, 'i-amphtml-layout-awaiting-size');
       break;
--- a/packages/optimizer/lib/transformers/ApplyLayout.js
+++ b/packages/optimizer/lib/transformers/ApplyLayout.js
@@ -81,7 +81,7 @@ function apply(layout, width, height, node) {
       // Do nothing here.
       break;
     case 'fluid':
-      styles += 'height:0;';
+      styles += 'width:100%;height:0px;';
       addClass(node, 'i-amphtml-layout-awaiting-size');
       break;
     case 'flex-item':
--- a/packages/optimizer/lib/transformers/ApplyLayout.js
+++ b/packages/optimizer/lib/transformers/ApplyLayout.js
@@ -36,6 +36,7 @@ const SUPPORTED_LAYOUTS = [
   'container',
   'fill',
   'flex-item',
+  'fluid',
   'intrinsic',
 ];
 
@@ -78,6 +79,13 @@ function apply(layout, width, height, node) {
     case 'fill':
     case 'container':
       // Do nothing here.
+      break;
+    case 'fluid':
+      if (width.isSet) {
+        styles = `width:${width.numeral}${width.unit};`;
+      }
+      styles += 'height:0;';
+      addClass(node, 'i-amphtml-layout-awaiting-size');
       break;
     case 'flex-item':
       if (width.isSet) {
--- a/packages/optimizer/lib/transformers/ApplyLayout.js
+++ b/packages/optimizer/lib/transformers/ApplyLayout.js
@@ -81,7 +81,7 @@ function apply(layout, width, height, node) {
       // Do nothing here.
       break;
     case 'fluid':
-      styles += 'width:100%;height:0px;';
+      styles += 'width:100%;height:0;';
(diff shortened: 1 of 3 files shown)


packages/optimizer/lib/transformers/ApplyLayout.js
```diff
@@ -79,6 +80,13 @@ function apply(layout, width, height, node) {
     case 'container':
       // Do nothing here.
       break;
+    case 'fluid':
+      if (width.isSet) {
+        styles = `width:${width.numeral}${width.unit};`;
```
jridgewell, line 85: When looking into it, it seems `width` isn't actually used or respected by fluid ads. Instead, there's a `width: 100% !important` set by https://github.com/ampproject/amphtml/blob/0320666d03db37c4e0e363b55f828aa807ae321c/extensions/amp-ad/0.1/amp-ad.css#L76-L78

jridgewell (approved):

packages/optimizer/lib/transformers/ApplyLayout.js
westonruter: The Fluid docs [say](https://github.com/ampproject/amphtml/blob/main/extensions/amp-ad-network-doubleclick-impl/fluid.md):

> It will always occupy the maximum available width, and its height will be determined relative to that width.

But also:

> Note also that the width attribute is optional, and can be specified. When specified, the fluid creative will always occupy that width (unless used in conjunction with multi-size).

So it seem there is still a case where the `width` would be used?

jridgewell: I don't believe it can be, due to the `!important` override?

westonruter: Since the selector is for `amp-ad[data-a4a-upgrade-type='amp-ad-network-doubleclick-impl'][height='fluid']`, is the `data-a4a-upgrade-type="amp-ad-network-doubleclick-impl"` attribute always added at runtime?

jridgewell: Yup, doubleclick is the only thing that supports fluid, and that attr gets injected when `amp-ad` loads.

westonruter: Removed in d232ad24e5ab5a238843a5878df05b4bc39031dd

packages/optimizer/lib/transformers/ApplyLayout.js
```diff
@@ -79,6 +80,10 @@ function apply(layout, width, height, node) {
     case 'container':
       // Do nothing here.
       break;
+    case 'fluid':
+      styles += 'height:0;';
```
jridgewell, line 84: Nit: we had `width:100%;height:0px;` in https://github.com/ampproject/amphtml/pull/34560/files#diff-ec6474c38021deb5cc02df92e38f7ae44a0bbda93384581872671a9c1c882d6fR292

jridgewell (approved):

packages/optimizer/lib/transformers/ApplyLayout.js
westonruter: Added in d9876e2, but `0` seems better than `0px`

schlessera (changes requested):

schlessera (approved):

sebastianbenz (approved):
Thanks a lot!

PR #1235, opened 2021-05-21
Pull latest component versions from amphtml repo

sebastianbenz:
This will ensure that we'll always use the correct component version when auto importing scripts.


packages/optimizer/lib/fetchRuntimeParameters.js
```diff
@@ -88,30 +88,30 @@ async function initValidatorRules(runtimeParameters, customRuntimeParameters, co
...
-      customRuntimeParameters.bentoComponentInfo ||
-      config.bentoComponentInfo ||
-      (await fetchBentoComponentInfoFromCache_(config));
+    runtimeParameters.componentVersions =
+      customRuntimeParameters.componentVersions ||
+      config.componentVersions ||
+      (await fetchComponentVersionsFromCache_(config));
   } catch (error) {
-    config.log.error('Could not fetch bento component info');
+    config.log.error('Could not fetch latest component versions from amp.dev');
```
patrickkettner, line 96: Nit, if it can continue, it feels like it should be a warning rather than an error

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -141,12 +141,7 @@ class AutoExtensionImporter {
       this.log_.error('Missing validation rules, cannot auto import extensions');
       return;
     }
-    if (!this.bentoComponentInfo) {
-      this.bentoComponentInfo = {};
-      for (const bentoComponent of params.bentoComponentInfo || []) {
-        this.bentoComponentInfo[bentoComponent.name] = bentoComponent;
-      }
-    }
+    this.componentVersions = params.componentVersions || {};
```
patrickkettner, line 144: Since optional chaining is supported in node 14￼, Would you ever want to use it rather than this type of syntax? Or do you prefer this type of format￼

patrickkettner (approved):

packages/optimizer/lib/fetchRuntimeParameters.js
```diff
@@ -15,7 +15,7 @@
  */
 'mode strict';
 
-const URL_BENTO_COMPONENT_INFO = 'https://amp.dev/static/bento-components.json';
+const URL_COMPONENT_VERSIONS = 'https://amp.dev/static/files/component-versions.json';
```
westonruter, line 18: Does this correspond to the `latest` version of each component?

westonruter: It seems some extensions are missing, namely:


1. `amp-action-macro`
1. `amp-base-carousel` 
1. `amp-cache-url` (this lacks a `latest` in [the spec](https://github.com/ampproject/amphtml/blob/main/extensions/amp-cache-url/validator-amp-cache-url.protoascii))
1. `amp-google-assistant-assistjs`
1. `amp-mega-menu`
1. `amp-story-panning-media`
1. `amp-stream-gallery`
1. `amp-truncate-text`

westonruter: Also, it doesn't seem accurate. It has this:

```
"amp-lightbox-gallery": "1.0",
```

And yet 1.0 is not even a [valid version in the spec](https://github.com/ampproject/amphtml/blob/e3abe630d3e0bcce78115e0b51d5da776e427904/extensions/amp-lightbox-gallery/validator-amp-lightbox-gallery.protoascii#L22-L23).

westonruter: The invalid latest version appears to be an issue with:

1. `amp-facebook-comments` (has 1.0 but [should be 0.1](https://github.com/ampproject/amphtml/blob/e3abe630d3e0bcce78115e0b51d5da776e427904/extensions/amp-facebook-comments/validator-amp-facebook-comments.protoascii#L22-L23))
2. `amp-lightbox-gallery` (has 1.0 but [should be 0.1](https://github.com/ampproject/amphtml/blob/e3abe630d3e0bcce78115e0b51d5da776e427904/extensions/amp-lightbox-gallery/validator-amp-lightbox-gallery.protoascii#L22-L23))
3. `amp-twitter` (has 1.0 but [should be 0.1](https://github.com/ampproject/amphtml/blob/e3abe630d3e0bcce78115e0b51d5da776e427904/extensions/amp-twitter/validator-amp-twitter.protoascii#L22-L23))

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -205,31 +200,25 @@ class AutoExtensionImporter {
...
-    }
-    // Bento component version is not yet known to the validator
-    if (version != bentoComponent.version) {
-      return version;
+    // Get latest non-experimental prod version (as calculated based on experimental flag)
+    const latestProdVersion = this.componentVersions[extensionName];
+    // Also get latest version supported by the validator
+    const latestValidatorVersion = extension.version[extension.version.length - 1];
+    // Verify latest version based on validator support
+    if (latestValidatorVersion < latestProdVersion) {
```
westonruter, line 218: This could cause problems later as `0.11` should be a newer than `0.2`, but here it would be false. In other words, I think this should use [`semver`](https://www.npmjs.com/package/semver):

```suggestion
    if (semver.lt(latestValidatorVersion, latestProdVersion)) {
```

packages/optimizer/lib/fetchRuntimeParameters.js
westonruter: Also, it seems `bento-components.json` is not completely. Namely, it is missing these Bento components:

* [`amp-video-iframe`](https://github.com/ampproject/amphtml/blob/main/extensions/amp-video-iframe/1.0/amp-video-iframe.js) (v1.0)
* [`amp-vimeo`](https://github.com/ampproject/amphtml/blob/main/extensions/amp-vimeo/1.0/amp-vimeo.js) (v1.0)

rsimha: @westonruter Drive-by comment to suggest an alternative to https://amp.dev/static/bento-components.json:
- The source of truth for Bento components is in [this file](https://github.com/ampproject/amphtml/blob/main/build-system/compile/bundles.config.extensions.json) (search for `"npm": true`).
- These are the components that will be automatically published to npm (see [design doc](https://docs.google.com/document/d/1hUmJl-NkpujkJyVpRRARTeoXHSiSf3VvMm5ujv5D10A/edit)).
- In case it helps, we have a wrapper [node script](https://github.com/ampproject/amphtml/blob/main/build-system/npm-publish/get-extensions.js) that can extract the raw list of components.

/cc @estherkim @jridgewell @kristoferbaxter for visibility / additional comment.

sebastianbenz: @rsimha this is perfect! Exactly what I was looking for. Am I right, that `latestVerstion` will always point to the latest valid prod version?

Background: Bento components have been causing the problem here, but the real problem is that it's not possible to determine experimental components from the validation rules.

rsimha: > Am I right, that latestVerstion will always point to the latest valid prod version?

Yep!

packages/optimizer/lib/fetchRuntimeParameters.js
sebastianbenz: done

packages/optimizer/lib/transformers/AutoExtensionImporter.js
sebastianbenz: Good point, thankfully no longer needed with the updated endpoint.

packages/optimizer/lib/transformers/AutoExtensionImporter.js
sebastianbenz: How does optional chaining help here? Do you mean nullish coalescing operator?

packages/optimizer/lib/fetchRuntimeParameters.js
estherkim: Yes! Git tags are created as part of the nightly cuts.

sebastianbenz:
Thanks everyone for the feedback and help!

PR #1236, opened 2021-06-01
Update error handler to support module and nomodule version

sebastianbenz:


packages/optimizer/lib/transformers/AmpBoilerplateErrorHandler.js
```diff
@@ -48,7 +59,11 @@ class AmpBoilerplateErrorHandler {
     }
 
     const errorHandler = createElement('script', {'amp-onerror': ''});
-    insertText(errorHandler, ERROR_HANDLER_TRANSFORMED);
+    if (params.esmModuleEnabled) {
```
westonruter, line 62: This could alternatively check if `html` has the `transformed` attribute:

```suggestion
    if (html.attribs.transformed !== undefined) {
```

cf. https://github.com/ampproject/amp-toolbox-php/blob/929220681fd80891dc6f110607269ec7127d5a0b/src/Optimizer/Transformer/AmpBoilerplateErrorHandler.php#L62-L64

And yet https://github.com/ampproject/amp-toolbox-php/pull/218.

packages/optimizer/lib/transformers/RewriteAmpUrls.js
```diff
@@ -72,6 +72,7 @@ class RewriteAmpUrls {
     let node = head.firstChild;
     let referenceNode = findMetaViewport(head);
     const esm = this.esmModulesEnabled || params.esmModulesEnabled;
+    params.esmModulesEnabled = esm;
```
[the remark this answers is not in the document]
westonruter: Then again, it could just check if the `transformed` attribute is present on `html`.

packages/optimizer/lib/transformers/AmpBoilerplateErrorHandler.js
schlessera: I think we need to decouple the ESM/onError logic from the transformed state. The onError handler needs to be changed depending on whether ESM is enabled or not. This can happen regardless of transformed state.

packages/optimizer/spec/transformers/valid/AmpBoilerplateErrorHandler/adds_module_error_handler_when_boilerplate_present /input.html
```diff
@@ -0,0 +1,17 @@
+<!--
```
schlessera, line 1: There seems to be an issue with the directory name for this case.
- it has trailing whitespace right now
- it should contain a reference to ESM being enabled, as opposed to the already existing case with ESM off

schlessera (changes requested):

packages/optimizer/lib/transformers/AmpBoilerplateErrorHandler.js
westonruter: > The onError handler needs to be changed depending on whether ESM is enabled or not.

I don't think this is the case. See https://github.com/ampproject/amp-toolbox-php/pull/218#issuecomment-854998921.

schlessera: But why shouldn't you be allowed to use ES6 modules in a source AMP document that does not use the Optimizer...?

westonruter: That is, ESM depends on `transformed` also being present. So the new `amp-onerror` _could_ be conditioned by whether ESM is present regardless of `transformed`, but since ESM depends on `transformed` we might as well just use it as the signal to use the new `amp-onerror`.

westonruter: > But why shouldn't you be allowed to use ES6 modules in a source AMP document that does not use the Optimizer...?

I think that should be allowed, but currently it is not. The use of `v0.mjs` depends on `transformed`: https://github.com/ampproject/amphtml/blob/65a6e3a3c0fd6837858a5600f0ee62a22ba4ee68/validator/validator-main.protoascii#L3150-L3174

sebastianbenz: Weston is right, but there's still the theoretical option of a user explicitly disabling mjs for transformed AMP.

packages/optimizer/lib/transformers/RewriteAmpUrls.js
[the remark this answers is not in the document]
sebastianbenz: This just ensures that the parameter is set to avoid duplicating the initialization logic.

packages/optimizer/spec/transformers/valid/AmpBoilerplateErrorHandler/adds_error_handler_when_boilerplate_present_esm_enabled/input.html
```diff
@@ -1,3 +1,8 @@
+<!--
+{
+  "esmModuleEnabled": true
+}
+-->
 <!doctype html>
 <html ⚡>
 <head>
```
schlessera, line 8: Can we actually use the `.mjs` runtime in this input file? I know that this is not really needed here, but it will probably be needed in the PHP version which also uses these tests.

(could not flag the right line, this file uses the `.js` version a few lines below)

schlessera (changes requested):

packages/optimizer/spec/transformers/valid/AmpBoilerplateErrorHandler/adds_error_handler_when_boilerplate_present_esm_enabled/input.html
patrickkettner: out of curiosity, any reason we wouldn't do

```html
<script type="module" src="foo.mjs"></script>
<script nomodule src="foo.js"></script>
```

is that not supported by the php infra?

schlessera: Yes, it is, and with my current version of the code, it is also what is expected to determine the right error handler to use. That's why I'm suggesting using that instead of the regular `v0.js` version of the runtime here.

schlessera: It should be `v0.mjs` though, not `foo.mjs`, as the runtime is the only thing that can be reliably checked.

schlessera: Here's what I have in PHP so far: https://github.com/ampproject/amp-toolbox-php/pull/218/files
As you can see I'm detecting the `v0.mjs` script using an XPath query to decide which version of the errorhandler to use.

The current test in here (which I'd also run off of in PHP) claims that ESM is in use, but does not actually use it. That will break the detection I'm doing of course.

I do think this detection is currently the most correct way, as you could potentially have a user hand-code the ESM version with the configuration `esmModuleEnabled` still being `false`. In that case, the wrong errorhandler will be used in the JS version.

sebastianbenz: 👍 done

schlessera (approved):

packages/optimizer/spec/transformers/valid/AmpBoilerplateErrorHandler/adds_error_handler_when_boilerplate_present_esm_enabled/input.html
schlessera: Thanks!

westonruter (approved):

sebastianbenz:
Thanks everyone!

PR #1237, opened 2021-06-10
Make ESM config consistent

schlessera:
There are two ESM config settings floating around now: `esmModulesEnabled` (which was used from the beginning) and `esmModuleEnabled` (which was introduced in a recent PR re. the boilerplate error handler).

This PR makes the config consistent again by using the first version across the code.


sebastianbenz (approved):
Good catch! Thanks!

sebastianbenz:
I'll fix the tests in a different PR

PR #1239, opened 2021-06-14
fix docker image: transform environment variable values to boolean when possible

mathieu-pousse:
Hello, 

Here is a fix for the docker image and its environment variables handling.

As the variables are strings, the checks in the optimizer fail when they are comparing 'true' with true.

The fix transform the string to a boolean value when possible.

This fixes the following command: 

```
docker run -p 3000:3000 -e AMP_OPTIMIZER_MARKDOWN=false -e AMP_OPTIMIZER_SEPARATE_KEYFRAMES=false ampproject/toolbox-optimizer
```

Let me know

### DIFF
--- a/packages/optimizer-docker/utils.js
+++ b/packages/optimizer-docker/utils.js
@@ -4,6 +4,10 @@ function objectFromEntries(arr) {
   return Object.assign({}, ...Array.from(arr, ([k, v]) => ({[k]: v})));
 }
 
+function toBooleanIfPossible(value) {
+  return value === 'true' ? true : value === 'false' ? false : value;
+}
+
 /*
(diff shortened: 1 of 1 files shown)


packages/optimizer-docker/utils.js
```diff
@@ -4,6 +4,10 @@ function objectFromEntries(arr) {
   return Object.assign({}, ...Array.from(arr, ([k, v]) => ({[k]: v})));
 }
 
+function toBooleanIfPossible(value) {
```
samouri, line 7: nit: I'd prefer the fn name `maybeParseBool`.

samouri (approved):
This looks great. Thank you for the fix!

PR #1241, opened 2021-06-16
AutoExtensionImporter: Append extension scripts after runtime script

schlessera:
The current implementation of the `AutoExtensionImporter` transformer adds the needed extension scripts after the viewport node.

However, this creates markup where the head contains all extension scripts first, then the boilerplate style, and then the runtime. In terms of priorities, this seems backwards. Why load the extensions if the runtime itself is not yet registered?

This PR changes the transformer so that it first tries to append the extension scripts after an existing runtime script, and only falls back to using the viewport as a reference node if no runtime script was found.

**before:**
```
<meta viewport>
<script extension 1>
<script extension 2>
<script extension 3>
...
<style boilerplate>
<script runtime>
```

**after:**
```
<meta viewport>
<style boilerplate>
<script runtime>
<script extension 1>
<script extension 2>
<script extension 3>
...
```


sebastianbenz (approved):
This is of course much better! We've been relying on the ReorderHeadTransformer to put things right later. Thanks!

westonruter:
> first tries to append the extension scripts after an existing runtime script

Aside: As in https://github.com/ampproject/amp-wp/issues/6182#issuecomment-859961308, I'd like to explore appending scripts which are not render critical to the end of the `body` (e.g. `amp-lightbox` and `amp-analytics`).

PR #1248, opened 2021-06-26
Disable preloading runtime script/style by default

westonruter:
See https://github.com/ampproject/amp-toolbox-php/issues/17.

cf. https://github.com/ampproject/amp-toolbox-php/pull/261


packages/optimizer/README.md
```diff
@@ -333,7 +334,11 @@ Hero images are optimized by server-side rendering the `img` element inside the
 
 #### `preloadHeroImage`
 
-Deprecated, use [optimizeHeroImages](#optimizeheroimages) instead. 
+Deprecated, use [optimizeHeroImages](#optimizeheroimages) instead.
+
+#### `preloadEnabled`
```
westonruter, line 339: Maybe `runtimePreloadEnabled` would be better?

packages/optimizer/spec/transformers/valid/RewriteAmpUrls/adds_preloads/input.html
```diff
@@ -1,3 +1,8 @@
+<!--
+{
+  "preloadEnabled": true
```
westonruter, line 3: This doesn't seem to be getting recognized.

packages/optimizer/spec/transformers/valid/RewriteAmpUrls/adds_esm_and_preloads/input.html
```diff
@@ -0,0 +1,18 @@
+<!--
+{
+  "esmModulesEnabled": true,
+  "preloadEnabled": true
```
westonruter, line 4: This isn't getting recognized. Why?

packages/optimizer/lib/transformers/RewriteAmpUrls.js
```diff
@@ -60,6 +60,7 @@ const {calculateHost} = require('../RuntimeHostHelper');
 class RewriteAmpUrls {
   constructor(config) {
     this.esmModulesEnabled = config.esmModulesEnabled !== false;
+    this.preloadEnabled = config.preloadEnabled === true;
```
westonruter, line 63: Should this only be honored if the boilerplate was removed?

schlessera:
I would suggest implementing this option in the following way:

**`preloadRuntime`** (`string`):
- `always` => always preload the runtime script
- `never` => never preload the runtime script
- `auto` (default) => preload the runtime script iff the boilerplate is still present

packages/optimizer/spec/transformers/valid/RewriteAmpUrls/adds_esm_and_preloads/input.html
schlessera: This is passed into the transformer's `transform()` method as the second `params` argument and thus needs to be read from there. You can see that `params.esmModulesEnabled` is already being processed in that method.

packages/optimizer/lib/transformers/RewriteAmpUrls.js
sebastianbenz: Yes, only if boilertplate has been removed. I'm wondering whether we should make it configurable at all. I'd tend to remove the option.

westonruter:
I've removed the option to control whether preload links are added and now made it conditional based on whether the boilerplate is present.

packages/optimizer/lib/transformers/RewriteAmpUrls.js
westonruter: Done in 17cd9bd

sebastianbenz (approved):
Thanks!

PR #1251, opened 2021-07-05
Avoid duplicate auto generated ids

### LINKED ISSUE #1250: AMP optimizer removing media queries
We have noticed that media queries are stripped from our amp-images on amp optimized pages using the Node amp-optimizer.

However, the same exact page optimized on the google amp optimizer works fine with media queries intact. 

Here is an example of our amp-optimized page with media queries stripped from the images:
https://www.bedbathandbeyond.com/amp/wm-optimized/store/category/bath/13432/

Screen Shot 2021-07-02 at 11 10 41 AM

Google Cache optimized page
https://www.google.com/amp/s/www.bedbathandbeyond.com/amp/store/category/bath/13432/

Screen Shot 2021-07-02 at 11 11 38 AM

Non-optimized amp page
https://www.bedbathandbeyond.com/amp/store/category/bath/13432/

Screenshot of the media queries intact on the google cached version
Screen Shot 2021-07-02 at 11 14 37 AM

Screenshot of the media queries stripped from the images on the Node amp-optimizer version
Screen Shot 2021-07-02 at 11 17 15 AM

Details:
Amp Version: Version 2106182132000
Amp toolbox optimize Version: v2.8.0
We are using [default](https://github.com/ampproject/amp-toolbox/tree/master/packages/optimizer#readme) settings
We are not defining a function for imageOptimizer

sebastianbenz:
Fixes a problem in which the same auto-generated id is assigned to
multiple elements.

Fixes #1250

### DIFF
--- a/packages/optimizer/lib/transformers/ApplyCommonAttributes.js
+++ b/packages/optimizer/lib/transformers/ApplyCommonAttributes.js
@@ -164,13 +164,11 @@ class ApplyCommonAttributes {
     this.log = log;
     this.canRemoveBoilerplate = true;
     // node counter for id generation
-    this.counter = 0;
+    this.transformedNodesCounter = 0;
     // nodes to check for attributes
     this.nodesToTransform = [];
     // existing ids in the document
     this.ids = new Set();
-    // nodes that have been transformed
-    this.transformedNodes = [];
     this.attributeTransformations = {
       media: new MediaTransformer(),
       sizes: new SizesTransformer(),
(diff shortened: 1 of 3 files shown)


packages/optimizer/lib/transformers/ApplyCommonAttributes.js
```diff
@@ -264,8 +260,8 @@ class ApplyCommonAttributes {
       return node.attribs.id;
     }
     node.attribs = node.attribs || [];
-    const id = ID_PREFIX + this.counter;
-    this.counter++;
+    const id = ID_PREFIX + this.transformedNodesCounter;
```
schlessera, line 263: The "next ID index" and the "number of nodes transformed by this transformer seems completely incidental to me and will only work correctly iff:
- no other transformer uses IDs with the same prefix;
- no preexisting IDs with the same prefix are found in the source document.

A more robust solution would be to have a separate object responsible for handing out IDs (that can be reused across transformers) and to check the next available index before using it.

I've solved this in the following way in the PHP version: https://github.com/ampproject/amp-toolbox-php/blob/v0.6.0/src/Dom/Document.php#L1806-L1849

schlessera: Note that in the PHP version, this is not a separate object (yet), but it is located in the `Dom\Document` abstraction, instead of within a transformer.

sebastianbenz: I'm checking for existing ids here: https://github.com/ampproject/amp-toolbox/blob/main/packages/optimizer/lib/transformers/ApplyCommonAttributes.js#L265. The counter will increase until there's a free id. 

Moving this into a separate service is definitely the best approach, but I'd only add this when we actually need it.

schlessera: Ah, yes, I see. You're collecting the IDs in a `Set` to do lookups. Indeed, this will need to be extracted out when it is needed in another place, but only when that is actually the case.

This made me look into the performance tradeoffs of my approach vs yours in terms of checking existing IDs, though. I've been using `getElementById()`, assuming that would be a sort of hashed lookup. But reading up on it, I'm not so sure anymore, as this is meant to work with multiple, freely definable and namespaceable ID attributes in XML.

I'll have to test whether it might be more efficient to keep an up-to-date lookup table like yours within the `Dom\Document` abstraction...

sebastianbenz: I went with the Set as we don't have the luxury of a `getElementById` API and we're already traversing over all nodes...

PR #1258, opened 2021-07-19
Add terser format options

ediamin:
This PR added two config option for terser in MinifyHtml optimizer. The goals of this PR are,

1. To keep the original quote style in output. Do not change single quote to double quote or vice versa. `quote_style: 3` makes sure the source quote style keeps remain the same in output.
2.  The default value of `wrap_func_args` is true, which wraps function expressions that are passed as arguments(callback), in parenthesis. `wrap_func_args: false` config solves this issue.

To support the new settings I've changed the quote style and remove the extra parenthesis in `minifies_inline_amp-script` spec test output.


schlessera (approved):

schlessera:
Additional note:
1. is needed to make the spec tests work for PHP as well, as the library we're using does not modify the quotes as the `terser` default does.
2. seems like a logical bugfix, as without it, the modification actually _adds_ bytes to the output.

sebastianbenz (approved):

sebastianbenz:
Thanks!

PR #1264, opened 2021-10-03
optimizer: Restore runtime-host path functionality

mdmower:
PR #855 introduced two regressions:

- It became impossible to specify a runtime-host with a path in the URL
  (e.g. https://example.com/amp-runtime)
- The --rtv flag led to inconsistent AMP script URLs because
  runtime-host dropped `/rtv/<rtv>`

One of the advertised features of self-hosting from
https://github.com/ampproject/amphtml/blob/main/docs/spec/amp-framework-hosting.md
is:

- serve AMP pages and the framework from the same host, potentially improving content delivery times.

For this to be possible, the runtime should be served from a path under
the same domain as the amp pages.


sebastianbenz (approved):
Thanks for fixing this Matt!

PR #1272, opened 2021-12-07
Handle removal of "latestVersion" from extensions bundle configuration

erwinmombay:
latestVersion was recently removed in https://github.com/ampproject/amphtml/commit/4df29d5cb8c054a42fafc8cfe0e07259c678754d#diff-e1e01796130015ee103b63a398890524b618b40f0dd857a788d1c14003ee490b so tests started to fail. Need to handle its removal on the auto importer.

also the previous configuration seemed inaccurate such as "latestVersion" was left at 0.1 for example even though there was already a 1.0 newer version entry in the config. (see amp-timeago, amp-accordion)


erwinmombay:
@sebastianbenz PTAL

erwinmombay:
cc @samouri

sebastianbenz (approved):
Nice - thanks for fixing!

packages/optimizer/lib/transformers/AutoExtensionImporter.js
```diff
@@ -141,7 +141,23 @@ class AutoExtensionImporter {
     if (!this.componentVersions) {
       this.componentVersions = {};
       for (const component of params.componentVersions) {
-        this.componentVersions[component.name] = component.latestVersion;
+        if (Array.isArray(component.version)) {
```
samouri, line 144: Is this meant to replicate the `latestVersion` behavior? If yes, it is incorrect AFAICT. At some point in the past latestVersion stopped correlating with the highest number

samouri:
I'm a bit confused by the goal of `AutoImport` plugin. It makes sense to me in the context of someone first writing an AMP Page and wanting a good DevEx as they do it (sort of like an external build_cleaner).

But its clear to me that folks shouldn't keep an empty `<head>` and rely on AutoImporter to guess the right version for them, because that would mean we can never change the version AutoImporter adds as we update to new versions (Bento)

schlessera:
@sebastianbenz, @erwinmombay,
This change now pull in experimental versions by default, instead of only stable versions. As an example, it now uses version 1.0 of `amp-timeago` instead of version 0.1. Version 1.0 is still marked as experimental: https://amp.dev/documentation/components/amp-timeago-v1.0/?format=websites

packages/optimizer/spec/transformers/valid/AutoExtensionImporter/uses-non-experimental-versions/expected_output.html
```diff
@@ -5,7 +5,7 @@
...
-  <script async src="https://cdn.ampproject.org/v0.js"></script><script async src="https://cdn.ampproject.org/v0/amp-timeago-0.1.js" custom-element="amp-timeago"></script><script async src="https://cdn.ampproject.org/v0/amp-youtube-0.1.js" custom-element="amp-youtube"></script><script async src="https://cdn.ampproject.org/v0/amp-date-display-0.1.js" custom-element="amp-date-display"></script><script async src="https://cdn.ampproject.org/v0/amp-mustache-0.2.js" custom-template="amp-mustache"></script>
+  <script async src="https://cdn.ampproject.org/v0.js"></script><script async src="https://cdn.ampproject.org/v0/amp-timeago-1.0.js" custom-element="amp-timeago"></script><script async src="https://cdn.ampproject.org/v0/amp-youtube-0.1.js" custom-element="amp-youtube"></script><script async src="https://cdn.ampproject.org/v0/amp-date-display-1.0.js" custom-element="amp-date-display"></script><script async src="https://cdn.ampproject.org/v0/amp-mustache-0.2.js" custom-template="amp-mustache"></script>
```
schlessera, line 8: This test should have caught the experimental version, but as the test was adapted instead, the issue was not caught.

sebastianbenz:
@schlessera good catch! We need to fix this!

westonruter:
I've asked for there to be a new reference to easily read which component versions are still experimental: https://github.com/ampproject/amphtml/issues/36749#issuecomment-996160595.

packages/optimizer/spec/transformers/valid/AutoExtensionImporter/uses-non-experimental-versions/expected_output.html
erwinmombay: @schlessera thanks for pointing that out. will fix

PR #1273, opened 2022-01-04
fix auto fetching of "latestVersion" configuration from amphtml repo

erwinmombay:
the previous fix which assumed that the highest version of an extension should be automatically used by the auto importer is incorrect. for example on some instances the newest version is not yet in stable and should not be used.

We fetch the frozen latest version configuration and merge it to generate an "extensionConfig.json" with a "latestVersion" field


erwinmombay:
@sebastianbenz what is the different between `fetchRuntimeParameters` and `downloadRuntimeParameters`? i wasn't sure if i should make this same change to `fetchRuntimeParameters`

packages/optimizer/build/downloadRuntimeParameters.js
```diff
@@ -42,11 +43,28 @@ const fetchRuntimeParameters = require('../lib/fetchRuntimeParameters');
 
   const extensionConfigUrl =
     'https://raw.githubusercontent.com/ampproject/amphtml/main/build-system/compile/bundles.config.extensions.json';
-  const response = await fetch(extensionConfigUrl);
-  if (!response.ok) {
-    throw new Error(`Failed downloading ${extensionConfigUrl} with status ${response.status}`);
+  const latestVersionsUrl = `https://raw.githubusercontent.com/ampproject/amphtml/main/build-system/compile/bundles.legacy-latest-versions.jsonc`;
+
+  const configResponse = await fetch(extensionConfigUrl);
```
sebastianbenz, line 48: Can we parallelize the two requests?

packages/optimizer/build/downloadRuntimeParameters.js
```diff
@@ -42,11 +43,28 @@ const fetchRuntimeParameters = require('../lib/fetchRuntimeParameters');
...
+    throw new Error(
+      `Failed fetching latest component versions from ${latestVersionsUrl} with status: ${latestVersionsConfigResponse.status}`
+    );
+  }
+  const extensionConfig = await configResponse.json();
+  const latestVersionsConfig = JSON5.parse(await latestVersionsConfigResponse.text());
+  // We add back the "latestVersion" field so that the auto importer
+  // code knows what "stable" version of the extension to use.
+  extensionConfig.forEach((entry) => {
+    entry['latestVersion'] = latestVersionsConfig[entry.name];
```
sebastianbenz, line 66: Nice!

sebastianbenz (approved):
Thanks for fixing! One nit, otherwise looks good.

packages/optimizer/build/downloadRuntimeParameters.js
erwinmombay: yes. ill make the change

samouri (approved):

packages/optimizer/build/downloadRuntimeParameters.js
```diff
@@ -42,11 +43,28 @@ const fetchRuntimeParameters = require('../lib/fetchRuntimeParameters');
 
   const extensionConfigUrl =
     'https://raw.githubusercontent.com/ampproject/amphtml/main/build-system/compile/bundles.config.extensions.json';
-  const response = await fetch(extensionConfigUrl);
-  if (!response.ok) {
-    throw new Error(`Failed downloading ${extensionConfigUrl} with status ${response.status}`);
+  const latestVersionsUrl = `https://raw.githubusercontent.com/ampproject/amphtml/main/build-system/compile/bundles.legacy-latest-versions.jsonc`;
```
westonruter, line 46: Cool. I've essentially taken this same approach in https://github.com/ampproject/amp-wp/pull/6803.

packages/optimizer/build/downloadRuntimeParameters.js
erwinmombay: done.

PR #1277, opened 2022-01-12
Bail early in `SeparateKeyframes` transformer

### LINKED ISSUE #1276: ERROR Install postcss
```
ERROR Install postcss to move keyframe animations into amp-keyframes via `npm install postcss postcss-safe-parser cssnano-simple`
```

I got this error while trying to use @ampproject/toolbox-optimizer@3.0.0-canary.2

@sswissspidy ran into the same issue

cc @sebastianbenz in case he knows what the issue is

swissspidy:
Don't warn about missing dependencies if the transformer was disabled via the config anyway.

Fixes #1276


sebastianbenz (approved):
Good one! Thanks for fixing!

PR #1279, opened 2022-01-12
Fix: add latest version at runtime

sebastianbenz:


erwinmombay (approved):

erwinmombay:
@sebastianbenz could we add additional tests for this? im not sure why this wasn't detected in our e2e

sebastianbenz:
How can we test this? The problem is that our implementation only tests against the prod version. The only thing I can think of is that we run periodic tests against the canary version.

sebastianbenz:
Merging as test failures are unrealted.

Issue #1302, opened 2022-02-18
SSR: Duplicated ids in rendered markup, if there are nodes with attributes 'media', 'heights' or 'sizes'


DK-Stern:
# Issue

On SSR generated amp pages are some nodes with same id, which breaks the logic of generated css rules with `display: block`, so some images wouldn't be displayed.

# Reason

Because of decreasing the counter `transformedNodesCounter` on following line, we've got on some nodes duplicated ids generated:

https://github.com/ampproject/amp-toolbox/blob/b367a2db78e1934e5c3e72727777e7c29d6a3b35/packages/optimizer/lib/transformers/ApplyCommonAttributes.js#L220

I cannot say why this happen, but we can ensure an easy fix to preventing of generating duplicated ids.

# Solutions

## First possible solution

One possible solution would be, to add the generated id to the Set of `ids` before line 269 to find collisions.

https://github.com/ampproject/amp-toolbox/blob/b367a2db78e1934e5c3e72727777e7c29d6a3b35/packages/optimizer/lib/transformers/ApplyCommonAttributes.js#L265-L270

That would look like this:

```javascript
[...]
    if (this.ids.has(id)) {
      // generate a new id if this one already exists
      return this.getOrCreateId(node);
    }
    this.ids.add(id);
    return id;
  }
````

## Second possible solution

A other possible solution would be, to remove the else condition where the counter `transformedNodesCounter` will be decreased:

https://github.com/ampproject/amp-toolbox/blob/b367a2db78e1934e5c3e72727777e7c29d6a3b35/packages/optimizer/lib/transformers/ApplyCommonAttributes.js#L219-L221

# PR

For the first solution i've created a PR: https://github.com/ampproject/amp-toolbox/pull/1303

PR #1303, opened 2022-02-18
Fix for issue #1302: "SSR: Duplicated ids in rendered markup, if there are nodes with attributes 'media', 'heights' or 'sizes'"

DK-Stern:
Fix for: https://github.com/ampproject/amp-toolbox/issues/1302

### DIFF
--- a/packages/optimizer/lib/transformers/ApplyCommonAttributes.js
+++ b/packages/optimizer/lib/transformers/ApplyCommonAttributes.js
@@ -216,6 +216,7 @@ class ApplyCommonAttributes {
       }
       if (nodeHasBeenTransformed) {
         node.attribs.id = id;
+        this.ids.add(id);
       } else {
         this.transformedNodesCounter--;
       }
@@ -266,7 +267,6 @@ class ApplyCommonAttributes {
       // generate a new id if this one already exists
       return this.getOrCreateId(node);
     }
-    this.ids.add(id);
     return id;
   }
 }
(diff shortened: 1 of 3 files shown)


sebastianbenz:
Thanks a lot for starting the work on this.

I haven't fully understood yet which cases cause the bug. Could you please add a test case in this [dir](https://github.com/ampproject/amp-toolbox/tree/main/packages/optimizer/spec/transformers/valid/ServerSideRendering) (see [instructions](https://github.com/ampproject/amp-toolbox/tree/main/packages/optimizer#development--testing)) that exposes the bug.

DK-Stern:
Ok i found the problem.

The reason for some duplicate ids are, that in function `apply()` the counter `transformedNodesCounter` gets decreased every time when `nodeHasBeenTransformed ` is `false`. 

https://github.com/ampproject/amp-toolbox/blob/26e113b3d22f5f31e3240cec47e213c7ea5fcc8e/packages/optimizer/lib/transformers/ApplyCommonAttributes.js#L220-L224

The reason for that is the increasing of counter `transformedNodesCounter` in function `getOrCreateId()`, but the counter will only be increased for nodes without id! 

https://github.com/ampproject/amp-toolbox/blob/26e113b3d22f5f31e3240cec47e213c7ea5fcc8e/packages/optimizer/lib/transformers/ApplyCommonAttributes.js#L261-L267

So every time when a node with own id will not be transformed (because there is no attribute media, heights or sizes), the counter will be decreased without it was increased before. In combination witch missing collision detection of own generated ids, duplicated ids got placed in resulted markup.

sebastianbenz (approved):
Good catch! Thanks for fixing!

Issue #1304, opened 2022-02-18
SSR: Failed transforming node with heights, if attributes has more then one media query


DK-Stern:
# Issue

A node fails transforming, if it has 'heights'-attribute which contains more than one media query.
Also the attribute got **removed** on rendering, if it had failed on transforming. 

## For example

```html
<amp-carousel
          width="1"
          height="1"
          heights="(min-width:768px) 56.25%, (min-width:576px) and (max-width:767px) 100%, 133%"
          layout="responsive"
          type="slides"
          role="region"
        >
[...]
</amp-carousel>
```

Will be rendered by amp framework, without any validation errors and with correct media queries. But on SSR with amp optimizer i get following error and the 'heights'-attribute will be removed on rendered amp page:

```javascript
AMP Optimizer ServerSideRendering Cannot remove boilerplate. Failed transforming heights="(min-width:768px) 56.25%, (min-width:576px) and (max-width:767px) 100%, 133%". Error: Invalid sizes definition '(min-width:768px) 56.25%, (min-width:576px) and (max-width:767px) 100%, 133%'
    at parseSizes (/.../node_modules/@ampproject/toolbox-optimizer/lib/parseSizes.js:63:15)
    at HeightsTransformer.transform (/.../node_modules/@ampproject/toolbox-optimizer/lib/transformers/ApplyCommonAttributes.js:127:21)
    at ApplyCommonAttributes.apply (/.../node_modules/@ampproject/toolbox-optimizer/lib/transformers/ApplyCommonAttributes.js:207:76)
    at ServerSideRendering.transform (/.../node_modules/@ampproject/toolbox-optimizer/lib/transformers/ServerSideRendering.js:105:27)
    at /.../node_modules/@ampproject/toolbox-optimizer/lib/DomTransformer.js:208:61
    at DomTransformer.doProfile (/.../node_modules/@ampproject/toolbox-optimizer/lib/DomTransformer.js:188:14)
    at DomTransformer.transformTree (/.../node_modules/@ampproject/toolbox-optimizer/lib/DomTransformer.js:208:18)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async DomTransformer.transform (/.../node_modules/@ampproject/toolbox-optimizer/lib/DomTransformer.js:178:7)
    at async DomTransformer.transformHtml (/.../node_modules/@ampproject/toolbox-optimizer/lib/DomTransformer.js:183:12)
```

# Reason

[...]

PR #1305, opened 2022-02-18
Fix for issue #1304: "SSR: Transforming of nodes with attribute 'heights' fails, if value contains multiple media queries"

DK-Stern:
Fix for issue https://github.com/ampproject/amp-toolbox/issues/1304


sebastianbenz (approved):
Thanks, this makes a lot of sense.

PR #1318, opened 2022-04-14
Add support for Opaque redirects

abdulhannanali:
This PR adds support for opaque redirects in @ampproject/cloudflare-optimizer-scripts

### The Problem

If the origin server serves a redirect the `fetch` script automatically follows the redirect and afterwards AMP manipulates the body, leaving no opportunity for there to be a browser navigation based on the redirect status code. 

This is a further concern, when there are Cross Origin navigations present based on the redirects.

### The Solution

~I have added support for fetch's opaque redirect mode which happens to be  the Cloudflare's default mode for the original incoming requests object as well. However, these are overridden and there's no current support for such redirects in Cloudflare.~

~This will let the browser decide what to do with the redirects and will not interrupt the navigation flow.~

The code in PR explicitly makes sure that redirect mode is manual and redirect based status codes are relegated back to browser.

### DIFF
--- a/packages/cloudflare-optimizer-scripts/src/index.js
+++ b/packages/cloudflare-optimizer-scripts/src/index.js
@@ -89,7 +89,9 @@ async function handleRequest(event, config) {
 
   const response = await fetch(request, { redirect: 'manual' });
 
-  if (response.type === 'opaque-redirect') {
+
+  // Redirect based statuses should be returned
+  if (response.status >= 300 && response.status <= 399) {
     return response;
   }
 
--- a/packages/cloudflare-optimizer-scripts/src/index.js
+++ b/packages/cloudflare-optimizer-scripts/src/index.js
@@ -75,7 +75,20 @@ async function handleRequest(event, config) {
 
   // Immediately return if not GET.
   if (request.method !== 'GET') {
-    return fetch(request, {redirect: 'manual'});
+    const fetchPromise = fetch(request, {redirect: 'manual'});
+
+    // For reverse proxy based responses, Location header needs to be rewritten in some cases
+    if (isReverseProxy(config)) {
+      const response = await fetchPromise;
+
+      if (response.status >= 300 && response.status <= 399) {
+        return rewriteLocationHeaderIfRequired(response, config);
+      }
+
+      return response;
+    }
+
+    return fetchPromise;
   }
 
   if (config.enableKVCache) {
@@ -92,7 +105,7 @@ async function handleRequest(event, config) {
   // Redirect based statuses should be returned
   if (response.status >= 300 && response.status <= 399) {
     if (config.proxy && config.proxy.origin) {
-      return rewriteLocationHeaderIfRequired(response, config)
+      return rewriteLocationHeaderIfRequired(response, config);
     }
 
     return response;
@@ -271,40 +284,45 @@ function resetOptimizerForTesting() {
   ampOptimizer = null;
 }
 
+/**
+ * @description Determine whether {Response} is Redirect style
+ * @param {Response} response
+ * @returns {boolean}
+ */
+function isRedirectResponse(response) {
+  return response.status >= 300 && response.status <= 399;
+}
 
 /**
  * @param {!Response} response
  * @param {!ConfigDef} config
  * @description Rewrites the location header if the request is coming from the proxied origin, designed to work with
  */
- function rewriteLocationHeaderIfRequired (response, config) {
-  const locationHeader = (
-       response.headers.get('location') 
-    || response.headers.get('Location')
-    || response.headers.get('LOCATION')
-  )
+function rewriteLocationHeaderIfRequired(response, config) {
+  const locationHeader =
+    response.headers.get('location') ||
+    response.headers.get('Location') ||
+    response.headers.get('LOCATION');
 
   if (!locationHeader) {
-    return response
+    return response;
   }
 
+  const parsedLocation = new URL(locationHeader);
 
-  const parsedLocation = new URL(locationHeader)
-  
   if (config.proxy.origin === parsedLocation.hostname) {
-    parsedLocation.hostname = config.proxy.worker    
-    const newHeaders = new Headers(response.headers)
-    newHeaders.set('Location', parsedLocation.toString())
-  
+    parsedLocation.hostname = config.proxy.worker;
+    const newHeaders = new Headers(response.headers);
+    newHeaders.set('Location', parsedLocation.toString());
+
     return new Response(response.body, {
       headers: newHeaders,
       status: response.status,
       statusText: response.statusText,
-    })
+    });
   }
 
-
-  return response
+  return response;
 }
 
 module.exports = {
--- a/packages/cloudflare-optimizer-scripts/src/index.js
+++ b/packages/cloudflare-optimizer-scripts/src/index.js
@@ -75,20 +75,14 @@ async function handleRequest(event, config) {
 
   // Immediately return if not GET.
   if (request.method !== 'GET') {
-    const fetchPromise = fetch(request, {redirect: 'manual'});
+    const response = await fetch(request, {redirect: 'manual'});
 
     // For reverse proxy based responses, Location header needs to be rewritten in some cases
-    if (isReverseProxy(config)) {
-      const response = await fetchPromise;
-
-      if (response.status >= 300 && response.status <= 399) {
-        return rewriteLocationHeaderIfRequired(response, config);
-      }
-
-      return response;
+    if (isReverseProxy(config) && isRedirect(response)) {
+      return rewriteRedirectHeader(response, config);
     }
 
-    return fetchPromise;
+    return response;
   }
 
   if (config.enableKVCache) {
@@ -103,9 +97,9 @@ async function handleRequest(event, config) {
   const response = await fetch(request, {redirect: 'manual'});
 
   // Redirect based statuses should be returned
-  if (response.status >= 300 && response.status <= 399) {
-    if (config.proxy && config.proxy.origin) {
-      return rewriteLocationHeaderIfRequired(response, config);
+  if (isRedirect(response)) {
+    if (isReverseProxy(config)) {
+      return rewriteRedirectHeader(response, config);
     }
 
     return response;
@@ -289,7 +283,7 @@ function resetOptimizerForTesting() {
  * @param {Response} response
  * @returns {boolean}
  */
-function isRedirectResponse(response) {
+function isRedirect(response) {
   return response.status >= 300 && response.status <= 399;
 }
 
@@ -298,11 +292,8 @@ function isRedirectResponse(response) {
  * @param {!ConfigDef} config
  * @description Rewrites the location header if the request is coming from the proxied origin, designed to work with
  */
-function maybeRewriteRedirectHeader(response, config) {
-  const locationHeader =
-    response.headers.get('location') ||
-    response.headers.get('Location') ||
-    response.headers.get('LOCATION');
+function rewriteRedirectHeader(response, config) {
+  const locationHeader = response.headers.get('location');
 
   if (!locationHeader) {
     return response;
@@ -313,7 +304,7 @@ function maybeRewriteRedirectHeader(response, config) {
   if (config.proxy.origin === parsedLocation.hostname) {
     parsedLocation.hostname = config.proxy.worker;
     const newHeaders = new Headers(response.headers);
-    newHeaders.set('Location', parsedLocation.toString());
+    newHeaders.set('location', parsedLocation.toString());
 
     return new Response(response.body, {
       headers: newHeaders,
--- a/packages/cloudflare-optimizer-scripts/src/index.js
+++ b/packages/cloudflare-optimizer-scripts/src/index.js
@@ -91,6 +91,10 @@ async function handleRequest(event, config) {
 
   // Redirect based statuses should be returned
   if (response.status >= 300 && response.status <= 399) {
+    if (config.proxy && config.proxy.origin) {
+      return rewriteLocationHeaderIfRequired(response, config)
+    }
+
     return response;
   }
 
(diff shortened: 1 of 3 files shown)


abdulhannanali:
Reproduced on https://opaque-cf-redirects-amp-optimizer-reproduction.hannan.codes/footballboots

The origin server tries a redirect to www.footballboots.co.uk but due to it being an AMP page, no redirect is happening.

abdulhannanali:
# Reproduction Live Demo

- Go to https://reproduction-github-amptoolbox-pr-1318.hannan.codes/amp 
- origin server wants to redirect it to amp.dev
- However, we are serve amp page on the same url

# Fix Live Demo

The fixed version https://reproduction-github-amptoolbox-pr-1318.hannan.codes/amp  will instead redirect to amp.dev

abdulhannanali:
So it turned out, my earlier approach relying on opaqueredirect isn't supported by Cloudflare Worker, instead they actually do make available the actual redirect stauts code in the redirect manual mode.

Moreover, the type property on Response object isn't supported at all by Cloudflare. Now, the code is returning the redirect response based on the status code, so the client can handle it

abdulhannanali:
Hi Maintainers,

I just wanted to add that this merge should also come with a notice for users to purge KV Cache since otherwise, the workers will keep serving in accurate content.

abdulhannanali:
@sebastianbenz @samouri 

Hi Friend! Could I get an ET for how much time a PR merge usually takes, I want to know as I have a client waiting on this to be merged?

abdulhannanali:
@samouri Hi, is it going to be accepted?

packages/cloudflare-optimizer-scripts/test/index.test.js
```diff
@@ -105,11 +105,36 @@ describe('handleEvent', () => {
...
+    const event = {
+      request: {method: 'GET', url: 'https://example.com/'},
+      respondWith: jest.fn(),
+      passThroughOnException: jest.fn(),
+    };
+
+    global.fetch.mockReturnValue(mockedResponse);
+    handleEvent(event, defaultConfig);
+
+    expect(global.fetch).toBeCalledTimes(1);
```
samouri, line 123: nit: testing implementation details is more brittle than testing function output of `respondWith`. In this case you are already checking for referentially equal returned value of the output. Does testing the number of times fetch was called or that it included a manual redirect directive increase confidence? Can we test for those features in an externally visible way instead?

samouri (commented):
Sorry for the delay! I missed this in my GH Notifications.
Thank you so much for this PR. It looks great.

One blocking question, the rest are nits

packages/cloudflare-optimizer-scripts/test/index.test.js
```diff
@@ -74,7 +74,7 @@ describe('handleEvent', () => {
     const input = `<html amp><body></body></html>`;
     const incomingResponse = getResponse(input);
     global.fetch.mockReturnValue(incomingResponse);
-    AmpOptimizer.transformHtmlSpy.mockReturnValue(Promise.reject('Fail.'));
+    AmpOptimizer.transformHtmlSpy.mockImplementation(() => Promise.reject('Fail.'));
```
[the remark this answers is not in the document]
abdulhannanali: @samouri By calling Promise.reject in the main test, the test fails, this is avoided with mockImplementation, which invokes this rejection only once transformHTMLSpy is invoked.

packages/cloudflare-optimizer-scripts/src/index.js
```diff
@@ -85,7 +87,14 @@ async function handleRequest(event, config) {
     }
   }
 
-  const response = await fetch(request);
+  const response = await fetch(request, { redirect: 'manual' });
+
+
+  // Redirect based statuses should be returned
```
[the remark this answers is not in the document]
abdulhannanali: @samouri 

That's certainly a question I had for the AMP team

The reason I choose this approach was 

- to stay on the safe side, as this lets the default behavior to persist and client can handle the reaction to redirects based on it's preferences

- We'd need another outgoing fetch request within the worker with redirect mode set to follow to determine the final origin

A downside however, is that it requires extra worker invocation if the final page is still served on the same origin which costs a little extra. 


Do you think we can give the user the ability to chose between the options?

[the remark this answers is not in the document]
samouri: > We'd need another outgoing fetch request within the worker with redirect mode set to follow to determine the final origin

Why is that? Because there may be a series of redirects?
Is there an issue with returning the first redirect and allowing the followup redirects to happen in later requests?

> to stay on the safe side, as this lets the default behavior to persist and client can handle the reaction to redirects based on it's preferences

Which client? Are you saying allow for customizing this behavior via the [config](https://github.com/ampproject/amp-toolbox/blob/1abaa6b3f72e517701fb394dd7ef41575c17c2e5/packages/cloudflare-optimizer-scripts/src/index.js#L25-L34) parameter? More customization options is usually a good things, but need to make sure the default makes sense.

----


I'm not sure what the two options are? Here's the situation I'm thinking about:

1. Browser requests `worker.com/example`
2. CF Worker has its `handleEvent` called and attempts to fetch AMP Document from `origin.com/example`
3. Origin returns  a redirect with Location `origin.com/example-new`
4. If CF Worker returns the Response without rewriting the header, the Browser will now redirect to the Origin! This violates expectations. The CF Worker should rewrite the Location header to redirect to `worker.com/example-new` instead.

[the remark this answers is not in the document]
abdulhannanali: Hi @samouri 

Apologies for the delay in response. 

> I'm not sure what the two options are? Here's the situation I'm thinking about:
> Browser requests worker.com/example
> CF Worker has its handleEvent called and attempts to fetch AMP Document from origin.com/example
> Origin returns a redirect with header Location: origin.com/example-new
> If CF Worker returns the Response without rewriting the header, the Browser will now redirect to the Origin! This violates expectations. The CF Worker should rewrite the Location header to redirect to worker.com/example-new instead.

Thank you very much for providing this example and shedding the light on this case. I was thinking more in terms of the cross-origin redirects where it's not an issue but I get it now. Moreover, this can certainly be even a larger concern in the Proxy mode too, if there's nothing in between the browser.

> Why is that? Because there may be a series of redirects?

> Is there an issue with the most naive approach of returning the first redirect and allowing the followup redirects to happen in later requests?

Since I didn't have the **location rewrite case from origin to worker** in mind, I understood initial question more in terms of intermediary requests being skipped if the final page, happens to be same origin when there are multiple redirects. 

The options I was suggesting were

- Skip the intermediary redirects where possible. 
- Let the Browser itself take care of intermediary redirects

Do you think it's a feasible option or something to be considered, given that it can increase the Egress from the Cloudflare Worker as well as may impact latency and increase complexity of the worker?

packages/cloudflare-optimizer-scripts/test/index.test.js
abdulhannanali: Thank you for detailed review. I have made some changes and removed the brittle fetch part.

I think as long as event.respondWith is responding with a same mocked redirect response we are good.

packages/cloudflare-optimizer-scripts/src/index.js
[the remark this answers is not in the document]
abdulhannanali: I will make a PR with Location rewrite in the morning. Have a great week ahead. 👍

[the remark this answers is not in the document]
samouri: Just push whenever this is ready and I'll re-review :)

abdulhannanali:
@sebastianbenz Apologies for not responding earlier, I will take a look tonight. I had my mid-term exams going on so didn't have much time on my hands.

packages/cloudflare-optimizer-scripts/src/index.js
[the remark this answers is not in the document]
abdulhannanali: Hi @samouri, 

A big apologies for late response, first of all, I will make sure that any future PRs are on time and such delay doesn't occur again. 

---

Now, coming to the PR,

I have made changes, with regards to 

- Rewriting Location Header for Redirects (both GET and non-GET) 
- The above rewriting logic is done only in Proxy mode 

Since in the Cloudlfare Routes mode, Cloudflare itself is intercepting the requests, the redirects should be intercepted as well, so I only implemented this functionality for Proxy mode. 

Can you think of any valid cases where Routes mode might require Location rewriting as well?

I can implement it, if there are any cases that you can think of. In that case, should we read the hostname from the request body, or require the user to give an origin in configuration regardless of whether it's Proxy or Route mode? 


---- 

Thank you so much for reviewing the PR and spending your valuable time reviewing the PR. :heart: 
 
--- 

Looking forward to make more PRs to this repository in future without the horrendous delays

packages/cloudflare-optimizer-scripts/test/index.test.js
```diff
@@ -105,11 +129,62 @@ describe('handleEvent', () => {
     expect(output).toBe(`transformed-${input}`);
   });
 
+  it('should handle redirects opaquely', async () => {
```
samouri, line 132:
```suggestion
  it('should passthrough opaque redirects unmodified', async () => {
```

packages/cloudflare-optimizer-scripts/test/index.test.js
```diff
@@ -105,11 +129,62 @@ describe('handleEvent', () => {
...
+      request: {method: 'GET', url: 'https://example.com/'},
+      respondWith: jest.fn(),
+      passThroughOnException: jest.fn(),
+    };
+
+    global.fetch.mockImplementation(() => Promise.resolve(mockedResponse));
+
+    handleEvent(event, defaultConfig);
+    expect(await event.respondWith.mock.calls[0][0]).toBe(mockedResponse);
+    expect(event.respondWith).toHaveBeenCalledTimes(1);
```
samouri, line 148: nit: not sure this is a useful assertion
```suggestion
```

packages/cloudflare-optimizer-scripts/test/index.test.js
```diff
@@ -105,11 +129,62 @@ describe('handleEvent', () => {
...
+  });
+
+  it('should rewrite location header for redirects to origin in proxy mode', async () => {
```
samouri, line 151: I think I'm confused -- how is this test different from the one on Line 74 with the title `should rewrite Location header for Redirect in response to Non-Get Requests (Proxy Mode)`?

samouri (commented):
My turn for the delay, as I was just on vacation 🏖️ 
Thank you for the updates! They are looking good.
I do agree that the Location rewrite only applies to the proxy use-case.

packages/cloudflare-optimizer-scripts/src/index.js
```diff
@@ -85,7 +100,17 @@ async function handleRequest(event, config) {
     }
   }
 
-  const response = await fetch(request);
+  const response = await fetch(request, {redirect: 'manual'});
+
+  // Redirect based statuses should be returned
+  if (response.status >= 300 && response.status <= 399) {
+    if (config.proxy && config.proxy.origin) {
```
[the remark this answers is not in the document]
abdulhannanali: I am using shouldRewriteRedirectHeader now

packages/cloudflare-optimizer-scripts/test/index.test.js
abdulhannanali: There are two usage sites, since non GET methods have different forwarding logic I am making sure that the Redirect headers are rerwriten in that case as well. 

Is that in our domain of responsibilities?

packages/cloudflare-optimizer-scripts/test/index.test.js
```diff
@@ -70,11 +71,31 @@ describe('handleEvent', () => {
     expect(await event.respondWith.mock.calls[0][0]).toBe(incomingResponse);
   });
 
+  it('should rewrite Location header for Redirect in response to Non-Get Requests (Proxy Mode)', async () => {
```
abdulhannanali, line 74: @samouri 

It's not duplication but since the worker has different logic handling Get and Non-Get scenarios. We need to account that both scenarios redirects are rewritten properly

packages/cloudflare-optimizer-scripts/src/index.js
```diff
@@ -85,7 +94,17 @@ async function handleRequest(event, config) {
     }
   }
 
-  const response = await fetch(request);
+  const response = await fetch(request, {redirect: 'manual'});
+
+  // Redirect based statuses should be returned
+  if (isRedirect(response)) {
```
abdulhannanali, line 100: @samouri Would this suffice or should I move this logic into rewriteRedirectHeader and rename it to `maybeRewriteRedirectHeader`  . In that case, would it be okay if based on the return value of the function we are either returning the response or continuing with the below lines of code? 

Just wanted to confirm the pattern you'll prefer.

packages/cloudflare-optimizer-scripts/src/index.js
```diff
@@ -259,6 +278,44 @@ function resetOptimizerForTesting() {
...
+  return response.status >= 300 && response.status <= 399;
+}
+
+/**
+ * @param {!Response} response
+ * @param {!ConfigDef} config
+ * @description Rewrites the location header if the request is coming from the proxied origin, designed to work with
+ */
+function rewriteRedirectHeader(response, config) {
+  const locationHeader = response.headers.get('location');
```
abdulhannanali, line 296: I am changing the cases of all the strings passed to `Header` to lowercase, since even though this won't matter in production. Since, we are using Map as polyfills the case-sensitivity matters there. 

I will replace this with a better poly fill later on.

abdulhannanali (commented):
Hi @samouri 

Thank you for providing the review. Hope your vacation went really well. I can't wait for summers myself, my college is still on going and we are seeing some of the hottest weather in two decades. 

Please ignore my last review as I think it was not very organized. Here's a cleaner one. 

I apologize for delay again. 

Yours Sincerely,
Hannan

packages/cloudflare-optimizer-scripts/src/index.js
samouri: Either way is fine with me! But consistency would be good.

Right now one of them is written as nested if conditions whereas the other combines the conditions with `&&` which is preferred in this case.

packages/cloudflare-optimizer-scripts/src/index.js
```diff
@@ -66,14 +66,23 @@ async function handleRequest(event, config) {
...
+
   request = new Request(url.toString(), request);
 
   // Immediately return if not GET.
   if (request.method !== 'GET') {
-    return fetch(request);
+    const response = await fetch(request, {redirect: 'manual'});
+
+    // For reverse proxy based responses, Location header needs to be rewritten in some cases
+    if (isReverseProxy(config) && isRedirect(response)) {
```
samouri, line 81: feel free to ignore, but technically the code would be faster if we separated the two checks. because the `fetch` doesn't ever need to be awaited here in the non-proxy case.

Something like:
```js
if (!isReverseProxy(config)) {
  return fetch(request);
}
const response = await fetch(request, {redirect: 'manual'});
return isRedirect(response) 
    ? rewriteRedirectHeader(response, config)
    : response
}
```

packages/cloudflare-optimizer-scripts/src/index.js
samouri: Sounds good!

samouri (approved):

samouri:
@abdulhannanali: thank you for your persistance! This PR LGTM now.

/to @jridgewell for merging/release

packages/cloudflare-optimizer-scripts/src/index.js
abdulhannanali: Thank you so much for all your guidance, This is a genius way to structure logic in order to optimize. 
I have made the change

PR #1319, opened 2022-04-18
Fix i-amphtml-sizer responsive issue

### LINKED ISSUE #1314: Responsive issue of i-amphtml-sizer created by the ServerSideRendering
Related to https://github.com/ampproject/amp-toolbox-php/issues/511

Consider an example snippet like this
```html
<amp-embed
  width="780"
  height="100"
  heights="(max-width:645px) 100%, (max-width:845px) 31%, 23%"
  layout="responsive"
  type="zergnet"
  data-zergid="42658"
></amp-embed>
```
When we use an element with `heights` attribute the SSR or more specifically the `createResponsiveSizer` method in [ApplyLayout.js](https://github.com/ampproject/amp-toolbox/blob/main/packages/optimizer/lib/transformers/ApplyLayout.js#L128) adds an inline `padding-top` in the sizer element. But the SSR also generates amp-custom styles like this,
```html
    <style amp-custom>
      #i-amp-0 > :first-child {
        padding-top: 23%;
      }
      @media (max-width: 845px) {
        #i-amp-0 > :first-child {
          padding-top: 31%;
        }
      }
      @media (max-width: 645px) {
        #i-amp-0 > :first-child {
          padding-top: 100%;
        }
      }
    </style>
    <amp-embed
      width="780"
      height="100"
      layout="responsive"
      type="zergnet"
      data-zergid="42658"
      id="i-amp-0"
      class="i-amphtml-layout-responsive i-amphtml-layout-size-defined"
      i-amphtml-layout="responsive"
      >
       <i-amphtml-sizer slot="i-amphtml-svc" style="display: block; padding-top: 12.8205%"></i-amphtml-sizer>
    </amp-embed>
```
For this output, the browser will always use the inline padding instead of the padding declared in amp-custom style. Hence, the generated markup is not responsive as expected. 

In addition to that, according to the [source code of amphtml](https://github.com/ampproject/amphtml/blob/main/src/custom-element.js#L1076), when we simply use the non-SSR element, it adds an inline padding-top using the `heights` attribute value. There is no affect of height/width ratio here. It only works if there is a `heights` attribute.

So, from my observation, we should remove the inline `padding-top` and only use the amp-custom style generated by the `heights` attribute value to get a responsive sizer.

ediamin:
Fixes #1314

This PR fixes the responsive issue of the `i-amphtml-sizer` generated by the SSR when we have a `heights` attribute.

### DIFF
--- a/packages/optimizer/lib/transformers/ApplyLayout.js
+++ b/packages/optimizer/lib/transformers/ApplyLayout.js
@@ -126,7 +126,9 @@ function createResponsiveSizer(node, width, height) {
 
   // Skip adding padding-top when heights attr is present. HeightsTransformer will take care of it.
   // https://github.com/ampproject/amp-toolbox/issues/1314
-  const style = !node.attribs.heights ? `;padding-top:${parseFloat(padding.toFixed(4))}%` : '';
+  const style = !hasAttribute(node, 'heights')
+    ? `;padding-top:${parseFloat(padding.toFixed(4))}%`
+    : '';
 
   const sizer = createElement('i-amphtml-sizer', {
     slot: 'i-amphtml-svc',
--- a/packages/optimizer/lib/transformers/ApplyLayout.js
+++ b/packages/optimizer/lib/transformers/ApplyLayout.js
@@ -111,7 +111,7 @@ function maybeAddSizerInto(node, layout, width, height) {
   }
   let sizer = null;
   if (layout === 'responsive') {
-    sizer = createResponsiveSizer();
+    sizer = createResponsiveSizer(node, width, height);
   } else if (layout === 'intrinsic') {
     sizer = createIntrinsicSizer(width, height);
   }
@@ -121,10 +121,13 @@ function maybeAddSizerInto(node, layout, width, height) {
   }
 }
 
-function createResponsiveSizer() {
+function createResponsiveSizer(node, width, height) {
+  const padding = (height.numeral / width.numeral) * 100;
+  const style = !node.attribs.heights ? `;padding-top:${parseFloat(padding.toFixed(4))}%` : '';
+
   const sizer = createElement('i-amphtml-sizer', {
     slot: 'i-amphtml-svc',
-    style: `display:block`,
+    style: `display:block${style}`,
   });
   return sizer;
 }
--- a/packages/optimizer/lib/transformers/ApplyLayout.js
+++ b/packages/optimizer/lib/transformers/ApplyLayout.js
@@ -123,6 +123,9 @@ function maybeAddSizerInto(node, layout, width, height) {
 
 function createResponsiveSizer(node, width, height) {
   const padding = (height.numeral / width.numeral) * 100;
+
+  // Skip adding padding-top when heights attr is present. HeightsTransformer will take care of it.
+  // https://github.com/ampproject/amp-toolbox/issues/1314
   const style = !node.attribs.heights ? `;padding-top:${parseFloat(padding.toFixed(4))}%` : '';
 
   const sizer = createElement('i-amphtml-sizer', {
--- a/packages/optimizer/lib/transformers/ApplyLayout.js
+++ b/packages/optimizer/lib/transformers/ApplyLayout.js
@@ -111,7 +111,7 @@ function maybeAddSizerInto(node, layout, width, height) {
   }
   let sizer = null;
   if (layout === 'responsive') {
-    sizer = createResponsiveSizer(width, height);
+    sizer = createResponsiveSizer();
   } else if (layout === 'intrinsic') {
     sizer = createIntrinsicSizer(width, height);
   }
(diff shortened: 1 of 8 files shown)


packages/optimizer/spec/transformers/valid/ServerSideRendering/generates_unique_ids_for_items_with_media_and_ignores_items_with_own_id/expected_output.html
```diff
@@ -6,13 +6,13 @@
...
-    <i-amphtml-sizer slot="i-amphtml-svc" style="display:block;padding-top:100%"></i-amphtml-sizer>
+    <i-amphtml-sizer slot="i-amphtml-svc" style="display:block"></i-amphtml-sizer>
   </amp-img>
   <amp-img id="customId" height="100" width="100" layout="responsive" src="img2.png" class="i-amphtml-layout-responsive i-amphtml-layout-size-defined" i-amphtml-layout="responsive">
-    <i-amphtml-sizer slot="i-amphtml-svc" style="display:block;padding-top:100%"></i-amphtml-sizer>
+    <i-amphtml-sizer slot="i-amphtml-svc" style="display:block"></i-amphtml-sizer>
   </amp-img>
   <amp-img height="100" width="100" layout="responsive" src="img3.png" class="i-amphtml-layout-responsive i-amphtml-layout-size-defined" i-amphtml-layout="responsive" id="i-amp-1">
-    <i-amphtml-sizer slot="i-amphtml-svc" style="display:block;padding-top:100%"></i-amphtml-sizer>
+    <i-amphtml-sizer slot="i-amphtml-svc" style="display:block"></i-amphtml-sizer>
```
westonruter, line 15: The elements in [`input.html`](https://github.com/ampproject/amp-toolbox/blob/337a494a293b672d5918dbe3bb8e6c868b150b03/packages/optimizer/spec/transformers/valid/ServerSideRendering/generates_unique_ids_for_items_with_media_and_ignores_items_with_own_id/input.html#L7-L9) do not have `heights` attributes. Is this change properly being limited to when `heights` is present?

ediamin: In [amphtml source](https://github.com/ampproject/amphtml/blob/main/src/core/static-layout.js#L155) I've found when the padding-top is being set. I've update the code and now padding-top will set when `heights` attribute is missing.

packages/optimizer/lib/transformers/ApplyLayout.js
```diff
@@ -121,11 +121,13 @@ function maybeAddSizerInto(node, layout, width, height) {
   }
 }
 
-function createResponsiveSizer(width, height) {
+function createResponsiveSizer(node, width, height) {
   const padding = (height.numeral / width.numeral) * 100;
+  const style = !node.attribs.heights ? `;padding-top:${parseFloat(padding.toFixed(4))}%` : '';
```
sebastianbenz, line 126: Can you please add a comment here and link to the issue.

sebastianbenz (approved):
LGTM - thanks for fixing! One nit.

packages/optimizer/lib/transformers/ApplyLayout.js
```diff
@@ -121,10 +121,13 @@ function maybeAddSizerInto(node, layout, width, height) {
   }
 }
 
-function createResponsiveSizer() {
+function createResponsiveSizer(node, width, height) {
+  const padding = (height.numeral / width.numeral) * 100;
+  const style = !node.attribs.heights ? `;padding-top:${parseFloat(padding.toFixed(4))}%` : '';
```
westonruter, line 129: Since we're already importing the `hasAttribute` function, might as well use it:

```suggestion
  const style = !hasAttribute(node, 'heights') ? `;padding-top:${parseFloat(padding.toFixed(4))}%` : '';
```

packages/optimizer/spec/transformers/valid/ServerSideRendering/generates_unique_ids_for_items_with_media_and_ignores_items_with_own_id/expected_output.html
westonruter: I don't see there in `static-layout.js` where the presence of the `heights` attribute determines whether `padding-top` is added. Maybe when the `heights` attribute is processed by the runtime, the `padding-top` later is removed/overridden? This appears so:

https://github.com/ampproject/amphtml/blob/160dcd9c93d34bb7da073a1240c287fc0cf14591/src/custom-element.js#L1068-L1078

ediamin: Sorry for the confusion. I wanted to say the same. Yes, you're right. [applyStaticLayout](https://github.com/ampproject/amphtml/blob/160dcd9c93d34bb7da073a1240c287fc0cf14591/src/custom-element.js#L1205) creates the sizer element first, then adds the padding-top based on the height and width attributes. Later it [gets overridden](https://github.com/ampproject/amphtml/blob/160dcd9c93d34bb7da073a1240c287fc0cf14591/src/custom-element.js#L1206) if the heights attribute is present.

westonruter (approved):
LGTM

sebastianbenz:
Thanks a lot!

PR #1339, opened 2022-10-10
CI: Only run GitHub workflows for ampproject repo

mdmower:
- Restrict CI workflow to ampproject/amp-toolbox repository
- Cleanup trailing whitespace


mdmower:
This is the same method used by the amphtml repository to limit CI runs to the ampproject. Example: https://github.com/ampproject/amphtml/blob/7c969ff5fd5406304f6a9867fa0db59b74e94524/.github/workflows/cut-nightly.yml#L17

sebastianbenz (approved):
Thanks!

PR #1358, opened 2024-05-23
amp optimizer to support ampjs.org (not enabled as default)

powerivq:
This PR will enable the transformer to recognize ampjs.org as valid AMP CDN. However, it does not automatically change the CDN to ampjs.org. The default is still cdn.ampproject.org. To use ampjs.org, can override with `ampUrlPrefix` param.

Adds test cases to ensure script tags pointing to either CDN get rewritten to the CDN configured.


danielrozenberg (approved):

danielrozenberg:
I think we would want to eventually change this so that it defaults to whatever CDN is used in the page being optimized, instead of defaulting to cdn.ampproject.org. wdyt?

powerivq:
@danielrozenberg I made it so that we only need to flip `DEFAULT_AMP_CACHE_HOST` to ampjs.org to achieve this. Once that is done, the transformer will change AMP docs containing either cdn.ampproject.org or ampjs.org into ampjs.org

PR #1359, opened 2024-05-23
Fix amp-toolbox PR build issues

powerivq:
lerna needs to be upgraded to work with newer nodejs version. Also change nodejs to 20-22.

Upgraded some other dependencies to fix various compatibility issues.


.github/workflows/CI.yml
```diff
@@ -11,19 +11,17 @@ jobs:
 
     strategy:
       matrix:
-        node: [12, 14, 16]
-        os: [ubuntu-latest, macOS-latest, windows-latest]
+        node: [16, 18, 20]
```
danielrozenberg, line 14: Officially we only support Node LTS, so this can just be `[20, 22]` instead

Does the Mac issue exist in those versions? We should keep it active, it's best to resolve this issue instead of removing support 😅

>

package.json
```diff
@@ -56,7 +56,7 @@
     "karma": "^6.4.0",
     "karma-chrome-launcher": "3.1.1",
     "karma-jasmine": "4.0.1",
-    "lerna": "5.1.4",
+    "lerna": "^7",
```
danielrozenberg, line 59: Please fix this to an exact version

powerivq: fixed.

.github/workflows/CI.yml
powerivq: Upgraded bunch of dependencies and it works now.

package.json
```diff
@@ -31,17 +31,17 @@
     "@types/debug": "4.1.5",
     "@types/diff": "5.0.0",
     "@types/escape-html": "1.0.0",
-    "@types/jest": "^26.0.21",
+    "@types/jest": "^29.5.12",
```
danielrozenberg, line 34: nit: pin to fixed version (no `^`)

danielrozenberg (approved):

